PR #193, opened 2026-05-27
feat(dsl): manifest-declared handler sources (ADR-072)

falese:
Extend LifecycleHookSchema with an optional `source` field that lets MFE
developers declare in the DSL manifest where each custom handler's
implementation lives — a relative path, a bare module (default export),
or `module#export`. Codegen emits a generated `handler-registry.ts` next
to `mfe.ts` that statically imports each sourced handler and wires it
into `super(manifest, { customHandlers: handlerRegistry })`. The
existing DI branch in `BaseMFE.invokeHandler` picks it up — no runtime
resolution code changes.

Hooks without `source` keep today's behaviour: a stub is stamped into
the generated `mfe.ts` and resolved via method reflection. Manifests
that don't declare any source are byte-identical to today (no registry
file emitted).

Scales identically across `RemoteMFE` (React) and `AngularRemoteMFE`
because the mechanism lives in BaseMFE + the shared codegen pipeline.
Both `mfe.ts.ejs` templates get the same conditional import + super()
arg; both template dirs get an identical `handler-registry.ts.ejs`.

The "Custom handler not found" error message now also suggests
declaring `source:` as the alternative discovery path.

ADR-072


falese:
## Testing with a real MFE runtime

`examples/abc-kids/flappy` is the easiest testbed — it already declares `onLoadBegin`/`onLoadComplete`/`onRenderBegin` etc. as inline-stub hooks. Pick one of those, move its body into a separate file, and let `source:` wire it back through the new generated registry.

### 1. Create a sourced handler

```bash
mkdir -p examples/abc-kids/flappy/src/platform/base-mfe/handlers
```

```ts
// examples/abc-kids/flappy/src/platform/base-mfe/handlers/onLoadBegin.ts
import type { Context } from '@seans-mfe-tool/runtime';

export async function onLoadBegin(context: Context): Promise<void> {
  console.log('[flappy][external-handler][onLoadBegin] sourced from a file ✓', {
    requestId: context.requestId,
    capability: context.capability,
    phase: context.phase,
  });
}
```

### 2. Point the manifest at it

In `examples/abc-kids/flappy/mfe-manifest.yaml`, edit the `Load → before` block:

```yaml
  - Load:
      type: platform
      lifecycle:
        before:
          - onLoadBegin:
              handler: onLoadBegin
              source: ./handlers/onLoadBegin    # NEW — ADR-072
              description: Log load lifecycle entry
```

The `source` path is relative to the generated `src/platform/base-mfe/mfe.ts`, so `./handlers/onLoadBegin` resolves to the file you just created.

### 3. Run codegen + build

```bash
cd examples/abc-kids/flappy
npm run build      # runs `seans-mfe-tool remote:generate && rspack build`
```

After codegen completes, verify the new artefacts:

```bash
cat src/platform/base-mfe/handler-registry.ts
# expect:  import { onLoadBegin } from './handlers/onLoadBegin';
#          export const handlerRegistry = { onLoadBegin: ... };

grep -A2 'constructor' src/platform/base-mfe/mfe.ts
# expect:  super(manifest, { customHandlers: handlerRegistry });

grep -c 'protected async onLoadBegin' src/platform/base-mfe/mfe.ts
# expect:  0   (stub suppressed — implementation lives in the external file)
```

### 4. Run it and watch the handler fire

```bash
npm run dev        # rspack serve on :3001 + mesh BFF on :4001
```

Open `http://localhost:3001/demo.html` and click the **load** button. In the browser console you should see:

```
[flappy][external-handler][onLoadBegin] sourced from a file ✓ { requestId: '...', capability: 'load', phase: 'before' }
```

That log line only fires if `BaseMFE.invokeHandler` resolved the name through `deps.customHandlers` (the generated registry) rather than the old method-reflection fallback — proving the full path: manifest → codegen → static import → DI map → runtime dispatch.

### 5. Try the module + named-export form

To exercise the `module#export` path, publish a tiny handler in a workspace package or pull it from `node_modules`. For a quick smoke test you can point at anything already installed:

```yaml
          - onLoadComplete:
              handler: onLoadComplete
              source: '@my-org/handlers#logComplete'
```

Codegen will emit `import { logComplete as onLoadComplete } from '@my-org/handlers';`. Build will fail at `rspack build` if the package isn't installed — that's by design, since this is the whole point of static-import resolution (build-time failures, not runtime).

### Back-compat sanity check

Revert your manifest edit — every existing example MFE should rebuild byte-identically to `main`. No `handler-registry.ts`, original `super(manifest)`, original stub on `mfe.ts`. If anything diverges for a no-`source` manifest, that's a regression and worth reporting on this PR.

---

falese:
## Live handler example: shell vs standalone detection

Just pushed `cdf06b7` — a real handler that uses the new `source:` mechanism. Sits in `examples/abc-kids/flappy/src/platform/base-mfe/handlers/detectExecutionContext.ts` and the flappy manifest pulls it in:

```yaml
  - Load:
      type: platform
      lifecycle:
        before:
          - detectExecutionContext:
              handler: detectExecutionContext
              source: ./handlers/detectExecutionContext
              description: Detect shell vs standalone via URL origin (ADR-072)
          - onLoadBegin:                     # ← inline-stub model preserved
              handler: onLoadBegin
              description: Log load lifecycle entry
```

### What it does

Compares `window.location.origin` to `new URL(context.inputs.remoteEntry).origin`. Writes the result onto the shared lifecycle context for downstream phases / the main capability body to read:

| Context key | Type | Value |
|---|---|---|
| `context.isShellContext` | `boolean` | `true` when loaded by a shell at a different origin; `false` when loaded standalone (`/demo.html` on the MFE's own port) |
| `context.executionContext` | `'shell' \| 'standalone' \| 'server'` | string label for logging / branching |
| `context.currentUrl` | `string` | `window.location.href` at detection time |

Fallback when `remoteEntry` isn't in context: path-based heuristic (`/demo.html` → standalone).

### Verifying the codegen path

```bash
cd examples/abc-kids/flappy
npm run build
```

Then inspect the regenerated platform files:

```bash
cat src/platform/base-mfe/handler-registry.ts
# import { detectExecutionContext } from './handlers/detectExecutionContext';
# import type { Context } from '@seans-mfe-tool/runtime';
# export const handlerRegistry = {
#   detectExecutionContext: detectExecutionContext as (context: Context) => Promise<unknown>,
# };

grep -A1 'constructor' src/platform/base-mfe/mfe.ts
# constructor(manifest: any) {
#   super(manifest, { customHandlers: handlerRegistry });

# Sourced hook → no stub. Inline hook → stub still emitted.
grep -c 'protected async detectExecutionContext' src/platform/base-mfe/mfe.ts   # → 0
grep -c 'protected async onLoadBegin'             src/platform/base-mfe/mfe.ts   # → 1
```

### Verifying the runtime path

#### Standalone

```bash
npm run dev    # rspack serve :3001 + mesh BFF :4001
# open http://localhost:3001/demo.html
# click "Test Load"
```

Console:
```
[flappy][detectExecutionContext] {
  currentUrl: 'http://localhost:3001/demo.html',
  currentOrigin: 'http://localhost:3001',
  mfeOrigin: 'http://localhost:3001',
  isShellContext: false,
  executionContext: 'standalone'
}
```

#### Shell

Spin up any shell that consumes `http://localhost:3001/remoteEntry.js` (the `examples/abc-kids/abc-kids-shell` is the obvious pick — its address is on a different port). Load the federated MFE; you should see:

```
[flappy][detectExecutionContext] {
  currentUrl: 'http://localhost:3000/...',     ← shell origin
  currentOrigin: 'http://localhost:3000',
  mfeOrigin: 'http://localhost:3001',          ← from remoteEntry input
  isShellContext: true,
  executionContext: 'shell'
}
```

If those two console logs come out as expected, the full chain works: manifest `source:` → generated `handler-registry.ts` → `customHandlers` DI map → `BaseMFE.invokeHandler` → external file's `detectExecutionContext()` → mutated `context` visible to every later hook and to the main `doLoad` body.

### Using the result downstream

Any later hook or capability override on the same load call can read it:

```ts
// e.g. in your mfe.ts override of doLoad, or in a later before/after hook
if (context.isShellContext) {
  // skip the auth dance, the shell already authenticated
} else {
  // standalone: bring up the dev login banner
}
```

Because the lifecycle context is shared across all phases of a capability execution (REQ-RUNTIME-002), no extra plumbing is needed.

---

falese:
Good catch — moved the handler out of the platform folder in `3806208`.

**New layout:**
```
examples/abc-kids/flappy/src/
├── platform/base-mfe/        # codegen-owned, regenerated each build
│   ├── mfe.ts                # imports handlerRegistry from sibling file
│   └── handler-registry.ts   # generated, imports '../../handlers/...'
├── handlers/                 # user-owned ✓
│   └── detectExecutionContext.ts
├── features/                 # domain capabilities
└── App.tsx
```

**Manifest** now spells the source relative to the generated `mfe.ts`:

```yaml
- detectExecutionContext:
    handler: detectExecutionContext
    source: ../../handlers/detectExecutionContext   # src/platform/base-mfe → src/handlers
```

ADR-072 updated with the convention so it's not lore: relative paths resolve from the generated `mfe.ts`, and `src/handlers/` is the recommended location. Everything else from the earlier recipe still works — just substitute the path. Sorry for the misleading first draft.

---

PR #310, opened 2026-07-25
feat(slots): close the app-code gap and make ADR-067 §4 real

falese:
> **#309 has merged; this is retargeted to `main` and merges clean.** The
> `mfe:validate` collision described below is resolved — the slot check ships as
> a rule inside #309's command.

## Why

A review of the slot system end to end — manifest → generated code → app code → registry rule → runtime — using `examples/meridian-station/meridian-console` as the reference provider.

**The manifest→codegen half is solid.** Grammar single-sourced (ADR-069), drift gate covering 21 MFEs, framework-agnostic template dispatch. Nothing needed fixing there.

**Everything downstream of it was unadopted or unimplemented.** Four findings, all verified against the code:

1. **`DeclaredSlot` had zero app-code usages repo-wide.** The generated component in every `slots.tsx`, plus `framework-react`'s and `framework-angular`'s published sugar, appeared only in their own tests and in codegen tests asserting they were emitted. Both real providers — `StationConsole` and `GameMenu` — hand-rolled `useCallback` + `slotContract.register(...)` instead. Not a correctness hole (`register()` calls `assertDeclared` internally), but the platform was generating an API nothing used, and every provider re-implemented the same four lines slightly differently. Three DOM conventions resulted: `data-declared-slot` on two regions, `data-berth` on the keyed ones, nothing at all in abc-kids.
2. **Nothing linked the manifest to app code at compile time.** `slots.tsx` exported the ids as *data*, so app code re-typed them as bare strings. Renaming a slot regenerated cleanly, kept `tsc` green, and broke only at runtime.
3. **ADR-067 §4's "registry-side validation target" was aspirational.** Both vendored registries contain zero occurrences of `"slot"`; `resolve.props` passed through opaquely. Rules lived as inline JSON inside `curl -d` heredocs, so nothing could read them and therefore nothing could check them. `registerMfe` compiled routes into closures without retaining them, so `GET /mfes` couldn't report placement targets even in principle.
4. **`schemas/dsl/manifest.schema.json` had zero occurrences of `providesSlots`** — stale since 2026-07-11, because `build:schema:dsl:check` existed in `package.json` but was never wired into CI.

None of these fail loudly at runtime: under ADR-066 an unbindable placement simply parks and waits forever.

### ADR-072 — `DeclaredSlot` is the app-code API, typed by the manifest

- Codegen emits `DeclaredSlotId`, a template-literal union derived from `providesSlots`. `{param}` segments become `${string}`, so a keyed id type-checks by interpolation: `` `berth.${berthId}` `` satisfies `` `berth.${string}` ``. No new naming scheme — the union *is* the constant set.
- `DeclaredSlot` now accepts standard element props and an `as` override, so it can replace the hand-rolled callbacks losslessly (inline `style` in StationConsole, semantic `<section aria-label>`/`<aside aria-label>` in GameMenu). The sugar was losing to the primitive partly because it was less capable than it.
- Both providers migrated. `slotContract.register()` stays public as the framework-adaptor escape hatch.
- The **fourth copy** of the slots template — inlined as a string literal in `examples/abc-kids/scripts/generate-games.mjs` — is removed; `slots.tsx` is generator-owned and the drift gate already keeps it current.

### ADR-073 — contract logic relocated, placement targets become validatable

`tsconfig` has path aliases for `@seans-mfe/{contracts,dsl,codegen,…}` but **none for the runtime package**, so CLI commands could not import the matcher. Re-implementing it design-time would have recreated exactly the two-sources drift ADR-069 had just removed. So the logic moved to the package both sides already reach:

- `createSlotContract` / `toProvidedSlotAddress` / types → `packages/contracts/src/slot-contract.ts`. `packages/runtime/src/slot-contract.ts` becomes a re-export shim, keeping the `@seans-mfe-tool/runtime` specifier every generated MFE imports. **The relocation is transparent: all 21 MFEs regenerate byte-identical.**
- New `createSlotAddressRegistry` answers the cross-MFE question a rule asks.
- **`slots-implemented`** joins #309's `ValidationRule` union in `packages/codegen/src/validate.ts` — see below.
- `slots:validate` — does every `resolve.props.slot` name an address some provider declares?
- `schemas/dsl/manifest.schema.json` regenerated; `build:schema:dsl:check` wired into CI.

### The `mfe:validate` collision with #309, and how it was resolved

@falese caught this in review: I checked that #296 was open but never checked whether a PR already implemented it. #309 did, and both PRs added `src/commands/mfe/validate.ts` and `schemas/mfe-validate.json` — add/add conflicts, plus four content conflicts.

**#309 is the host.** Its rules-based `validateMfeConsistency()` with a `ValidationRule` union and a thin I/O shell is the right shape for both questions, so the slot check became one more rule in that set rather than a second command wearing the same name:

- `slots-implemented` delegates to `findUnreferencedSlots` in `@seans-mfe/dsl`, so the matching (literal prefix before the first `{param}`) has one implementation.
- `MfeValidationInput` gains an optional `sources: SourceFile[]` — optional deliberately, so the function stays pure and usable without file IO, and the rule is skipped when sources are absent rather than flagging every declaration.
- Skipped too when the manifest declares no slots, so the 19 non-provider MFEs are unaffected.

The payoff of wiring it in rather than beside: **`check:mfe-consistency` now carries the slot rule across the whole fleet**, which my standalone command never did. ADR-073 §3 is corrected to describe this arrangement — an earlier draft had it backwards.

### Placement rules become data

Extracted verbatim into `examples/*/control-plane/rules.json` (7 registrations / 18 routes for Meridian; 14 / 40 for abc-kids, now emitted by the generator as data). Both register scripts became a thin `jq` loop, so what gets POSTed is exactly what is reviewed. CI runs `slots:validate` over both fleets.

### The demo registry enforces at rule-save

Registrations carry `providesSlots`; routes are retained as data with a new `GET /rules`; `POST /mfes` rejects a target a registered provider does not declare.

## Verification

Every new gate was verified **by making it fail**, since all four passed vacuously before:

| Gate | Negative test | Result |
|---|---|---|
| `DeclaredSlotId` | rename `main`→`primary`, regenerate | `TS2322: Type '"main"' is not assignable to type 'DeclaredSlotId'` |
| `slots-implemented` | declare a slot nothing registers | exit 1, `MFE_INCONSISTENT`, `✗ slots-implemented` |
| `slots:validate` | point a rule at `meridian-console/mian` | exit 1, names the declared alternatives |
| registry `POST /mfes` | same, against a **live** registry | HTTP 400; unknown provider warns+accepts; `root` accepts |

Standing gates, re-run after retargeting onto post-#309 `main`: lint **0 errors** · typecheck **clean** · **21 MFEs drift-free** · schemas committed · DSL schema current · **21 MFEs internally consistent** (slot rule evaluated on the two providers, skipped on the other 19) · both fleets' placement targets valid · meridian-console **15/15** and `tsc --noEmit` clean · **2125 tests pass**.

The `quality` job now runs four gates in one place: drift (#295), consistency (#296, carrying `slots-implemented`), DSL schema, and slot placement targets.

The migration's regression proof is that `StationConsole.test.tsx` and `GameMenu.test.tsx` — which assert the ids handed to `provideSlot` — **pass unchanged**.

All five mermaid diagrams in the new doc were parse-checked with mermaid.

## Two things reviewers should push back on if they disagree

- **The registry matcher is a copy, not a call.** Those registries are standalone dockerized services and `@seans-mfe/contracts` isn't published yet (MERGE-PLAN Phase 1). I first wrote ADR-073 §5 claiming a shared call, found that was wrong, and corrected it. `packages/contracts/src/__tests__/registry-slot-pin.test.ts` pins the copy against the real implementation across 17 cases including regex-metacharacter literality and keyed over-segmentation. Delete both when contracts ships.
- **`SlotTargetResult` discriminates on `status`, not a boolean `ok`.** This repo compiles without `strictNullChecks`, where TypeScript will not narrow a boolean-literal discriminant — verified empirically, not assumed.

## Not fixed, deliberately

- **5 pre-existing failures in `packages/oclif-base/src/__tests__/base-command.json-contract.test.ts`.** Confirmed identical on a clean tree via `git stash`. CI does not run them — `.github/workflows/test.yml` excludes `json-contract.test.ts` via `--testPathIgnorePatterns`, alongside four other suites — so they surface only in a local `npm run test:ci`.
- **`.tsx` tests never run.** Root `testMatch` is `*.test.[jt]s`, so `DeclaredSlot.test.tsx` and `MfeHost.test.tsx` have never executed. Enabling them needs React added to root devDeps, which ADR-050 governs — not a call to make silently in this PR. The additions there are correct specs but currently dormant; running coverage comes from the codegen tests and the example suites.

## Shared files touched

`package.json` (oclif topics `mfe`/`slots`, `@seans-mfe/contracts` dependency on codegen), `.github/workflows/test.yml`, `schemas/`, `CLAUDE.md`, `docs/spec.md` ADR index.

## Docs

New `docs/slot-architecture.md` — five diagrams (design time, build pipeline, rule authoring, runtime sequence, convergence) plus a runtime reference documenting six behaviors that existed only in the code, including the unqualified `'main'` default in `mountExperience` (host-owned, unrelated to any provider's declared `main`), `MAX_SLOT_ESCALATIONS`, and the per-remote-scope mount mutex that keyed fan-out depends on.

Also corrects `packages/dsl/src/schema.ts`, which cited ADR-057 for the `/` rejection where `slot-contract.ts` correctly cites ADR-068.

Refs #265
Refs #296

https://claude.ai/code/session_01Kam2PA1WzFarjVscA3JWYm


falese:
## Local-test runbook — accurate to `aa90fc9`

### 0. Build (required — `packages/runtime` changed)

`packages/runtime/src/slot-contract.ts` became a re-export shim, and `dist/runtime` is baked into the CLI image and staged into MFEs, so a plain `npm run build` is not optional here.

```bash
npm ci
npm run build                 # builds packages, dist/runtime, oclif manifest, schemas
npm run docker:build:cli      # only if you'll exercise the docker path
```

### 1. The four gates, verified by making each one fail

This is the heart of the PR — all four passed vacuously before, so passing proves nothing. Break each one:

**a) `DeclaredSlotId` — a manifest rename must fail the build**

```bash
sed -i 's/^  - id: main$/  - id: primary/' examples/meridian-station/meridian-console/mfe-manifest.yaml
npx ts-node scripts/check-mfe-drift.ts          # regenerates slots.tsx
cd examples/meridian-station/meridian-console && npm install && npx tsc --noEmit
```

Expect:
```
src/features/StationConsole/StationConsole.tsx(179,9): error TS2322:
  Type '"main"' is not assignable to type 'DeclaredSlotId'.
```

Restore with `git checkout examples/meridian-station/meridian-console/mfe-manifest.yaml && npx ts-node scripts/check-mfe-drift.ts`.

> On `main`, that same rename typechecks clean and only breaks in the browser. That contrast is the whole point of ADR-072.

**b) `mfe:validate` — a declared slot nothing implements**

```bash
./bin/run.js mfe:validate --dir examples/meridian-station/meridian-console   # ✓ 3/3
```

Now add `- id: ghost-rail` under `providesSlots` and re-run. Expect exit 1, `SLOT_DECLARED_BUT_UNREFERENCED`, and a note that the check is a static scan (it is a lint, not a proof — that limitation is stated in the command's own output).

**c) `slots:validate` — a rule aimed at a slot nobody declares**

```bash
./bin/run.js slots:validate \
  --rules examples/meridian-station/control-plane/rules.json \
  --manifests examples/meridian-station          # ✓ 18 routes
```

Then edit any `"slot": "meridian-console/main"` to `"meridian-console/mian"` in `rules.json`. Expect exit 1 naming the declared alternatives. `"meridian-console/berth.b1.extra"` is also rejected — a `{param}` matches exactly one segment.

**d) Registry rejects at rule-save (live service)**

```bash
cd examples/meridian-station/control-plane/registry && npm install
PORT=4500 node simple-registry.js &

cd ../.. && REGISTRY=http://localhost:4500 ./scripts/register-station.sh   # 7 MFEs, 18 routes

curl -s http://localhost:4500/rules | jq '.mfes[0]'    # NEW: routes are reportable

# undeclared target -> 400
curl -s -w '\nHTTP %{http_code}\n' -X POST http://localhost:4500/mfes \
  -H 'Content-Type: application/json' -d '{
    "registration":{"name":"rogue","version":"1.0.0","type":"remote","baseUrl":"http://localhost:9999"},
    "routes":[{"when":{"stateKey":"x"},"resolve":{"capability":"R","props":{"slot":"meridian-console/mian"}}}]}'
```

Expect `HTTP 400` with the declared vocabulary in the message. Swap the slot for `not-yet-registered/main` → `HTTP 200` plus a `⚠️ Registry:` warning on stderr (registration order isn't guaranteed, so unknown providers must not block), and `root` → `HTTP 200` (host-owned).

Kill with `pkill -f 'node simple-registry.js'`.

### 2. Standing gates

```bash
npm run lint                    # 0 errors (77 pre-existing warnings)
npm run typecheck               # clean
npm run build
git diff --exit-code schemas/
npm run check:mfe-drift:check   # 21 MFEs, no drift
npm run build:schema:dsl:check  # newly gated in CI
```

### 3. The regression proof for the migration

```bash
cd examples/meridian-station/meridian-console && npx jest    # 15/15, unchanged
```

These tests assert the ids handed to `provideSlot`. They were written against the hand-rolled ref callbacks and pass untouched against `DeclaredSlot` — that is the evidence the migration preserved behavior rather than just moving code.

### 4. Full end-to-end

```bash
cd examples/meridian-station && ./scripts/dev-up.sh && ./scripts/register-station.sh && ./scripts/console.sh
```

Correct result at `:5000`: the console with **six berth tiles** filled by the Angular `BerthTile` remote; selecting a domain composes into `meridian-console/main` and `meridian-console/status`. Daemon log shows `EXPERIENCE_RELAYED` per placement.

**If the tiles are empty**, check the DOM in this order:
1. `[data-declared-slot^="berth."]` — six of these means registration happened (this attribute is now emitted uniformly by `DeclaredSlot`; the e2e selector moved off the bespoke `[data-berth]`).
2. `[data-layout-slot="meridian-console/berth.b1"]` — means the host bound the address.

The first without the second means the desired-state entry never converged — look at `registerProvidedSlot` / `reconcileSlotNow` in `layout-manager.ts`, not at the component.

```bash
npx playwright test    # covers the selector change
```

### Known-red, unrelated

`npm run test:ci` exits non-zero: **5 pre-existing failures** in `packages/oclif-base/src/__tests__/base-command.json-contract.test.ts`. Confirmed identical on a clean tree (`git stash` → same 5). Not touched by this PR; 2106 tests pass.

---

docs/architecture-decisions/ADR-073-slot-contract-in-contracts-design-time-validation.md
```diff
@@ -0,0 +1,188 @@
...
+- An **unknown provider** is advisory (`unknown-provider`), not fatal. Registration
+  order is not guaranteed and a fleet may be validated one MFE at a time; treating
+  this as fatal would make the check unusable during incremental rollout.
+
+### 3. Two CLI surfaces, because there are two questions
+
+Both extend `BaseCommand`, implement `runCommand()`, use typed errors, and emit
+exactly one `CommandResult` line under `--json`.
+
+- **`mfe:validate [dir]`** — *intra*-MFE consistency: is every slot this MFE
```
falese: @claude i believe mfe:validate is on a seperate PR that hasn't landed, is this correct? if so can you determine the right strategy to merge

falese: You're right, and thanks for catching it — I checked `#296` was open but never checked whether a PR already implemented it. **#309** (`claude/issue-296-mfe-validate`, opened 2026-07-22, closes #296) implements `mfe:validate` and this PR collides with it head-on.

### The collision, measured

`git merge-tree origin/claude/issue-296-mfe-validate HEAD` — 6 conflicts, two of them **add/add**:

```
CONFLICT (add/add)   src/commands/mfe/validate.ts
CONFLICT (add/add)   schemas/mfe-validate.json
CONFLICT (content)   scripts/generate-schemas.ts
CONFLICT (content)   .github/workflows/test.yml
CONFLICT (content)   package.json
CONFLICT (content)   packages/codegen/src/index.ts
```

The textual conflicts are the small part. The real problem is semantic: **both PRs define a command called `mfe:validate` with different responsibilities.**

| | #309 | This PR |
|---|---|---|
| Asks | is this MFE's dependency/federation config consistent? | are this MFE's declared slots actually implemented? |
| Rules | `react-pinned`, `manifest-package-sync`, `shared-declared`, `shared-version-sync`, `runtime-declared`, `--typecheck` | `declared-but-unreferenced` |
| Pure core | `packages/codegen/src/validate.ts` | `packages/dsl/src/slot-validation.ts` |
| CI | `npm run check:mfe-consistency` | not CI-wired |

These aren't competing — they're the same question asked about two different things, and #309 already has the right shape for both: a `ValidationRule` union with a pure, I/O-free `validateMfeConsistency()` and a thin command shell. My slot check is one more rule in that set, not a second command wearing the same name.

Worth noting I wrote ADR-073 §3 saying *"this is the slot half of #296; the dependency assertions drop into the same command later."* That was right in principle and backwards in practice — #309 is the host, and the slot rule is what drops in.

### Recommended strategy: land #309 first, fold the slot rule into it

1. **Drop `src/commands/mfe/validate.ts` and `schemas/mfe-validate.json` from this PR.** Both add/add conflicts disappear; the four content conflicts left are one-line-each in shared files and resolve trivially. This PR keeps everything that doesn't collide and is the actual point: `DeclaredSlotId`, the `DeclaredSlot` migration, `slots:validate` (cross-MFE placement targets — zero overlap with #309), `rules.json`, registry rule-save enforcement, and the docs.
2. **Keep `findUnreferencedSlots` and its 7 unit tests** in `packages/dsl/src/slot-validation.ts`. Pure, no collision, and it is exactly the input #309's rule set needs — ready to wire with no rework.
3. **After #309 merges**, a small follow-up adds `slots-implemented` to its `ValidationRule` union: `validateMfeConsistency` gains an optional `sources: SourceFile[]` input, the command reads `src/**` (it already does directory I/O), and `check:mfe-consistency` picks the rule up for free across all 21 MFEs — better CI coverage than my standalone command had.
4. **ADR-073 §3 gets corrected** in this PR to describe that arrangement rather than a separate command.

Net effect: #309 lands unblocked and unchanged, this PR loses one command and no capability, and the slot check ends up better placed than where I put it — inside the fleet-wide gate instead of beside it.

The alternative — landing this first and making #309 rebase — costs Devin's PR a rework on the older, assigned, issue-closing branch to accommodate the newer one. I don't think that's the right trade, but it's your call on merge ordering.

Tell me which way and I'll push it.

---

falese: Done — pushed as `83043e6`. You picked "#309 first, wire the slot rule now", so this is now **stacked on #309**.

**Base retargeted to `claude/issue-296-mfe-validate`**, so the diff above shows only the slot work rather than re-displaying Devin's. GitHub will retarget to `main` automatically when #309 merges.

### What landed

**`584d90c` — merge, 6 conflicts resolved.** #309 wins `src/commands/mfe/validate.ts` and `schemas/mfe-validate.json` outright. `packages/codegen/src/index.ts`, `package.json`, and `.github/workflows/test.yml` take the union — the drift, DSL-schema, slot-placement, and MFE-consistency gates all survive. In `scripts/generate-schemas.ts` I kept #309's `mfe:validate` output shape and my `slots:validate`, and dropped my duplicate `mfe:validate` *input* entry, since #309's flags (`--typecheck`) are the ones that ship.

**`c1ff4bf` — `slots-implemented` as a rule.**

```
  ✓ react-pinned
  ✓ manifest-package-sync
  ✓ shared-declared
  ✓ shared-version-sync
  ✓ runtime-declared
  ✓ slots-implemented        ← new
```

- Delegates to `findUnreferencedSlots` in `@seans-mfe/dsl`, so the needle logic (literal prefix before the first `{param}`) has one implementation, not two.
- `MfeValidationInput` gains an **optional** `sources: SourceFile[]`. Optional on purpose: `validateMfeConsistency` stays pure and usable without file IO, and the rule is skipped when sources are absent rather than reporting every declaration as missing.
- Also skipped when the manifest declares no slots — the 19 non-provider MFEs are untouched.
- The command reads `src/**`, excluding the generated `slots.tsx`: it mirrors the manifest by construction, so counting it would make the check vacuous.

**`83043e6` — ADR-073 §3 corrected**, plus `CLAUDE.md` / `slot-architecture.md` / `slot-contract.md` pointing at the rule instead of the removed command.

### The part that turned out better than my original

`check:mfe-consistency` now carries the slot rule across the whole fleet — my standalone command was never CI-wired at all:

```
21 MFE(s) checked — all internally consistent.
```

`slots-implemented` evaluates on exactly the two slot providers (`abc-kids-home`, `meridian-console`) and is skipped on the other 19.

### Gates on the merged tree

lint **0 errors** · typecheck **clean** · **2125 tests pass** · 21 MFEs drift-free · schemas committed · DSL schema current · 21 MFEs internally consistent · both fleets' placement targets valid · meridian-console 15/15 and `tsc --noEmit` clean.

Negative test still fires, now inside the shared command:

```
  ✗ slots-implemented
      - Slot "ghost-rail" is declared in providesSlots but no source file
        references "ghost-rail". A placement targeting it would wait forever…
meridian-console is inconsistent: 1 problem(s).
    Code: MFE_INCONSISTENT   (exit 1)
```

One thing to flag: **#310 can't merge until #309 does.** If #309 stalls, say the word and I'll unstack — drop the merge commit and hold the slot rule as a follow-up, which puts this back to the option-1 shape.

---

falese:
## Runbook addendum — now accurate to `83043e6`

The [runbook above](https://github.com/falese/seans-mfe-tool/pull/310#issuecomment-5079267140) was written against `aa90fc9`, before this branch was stacked on #309. Everything in it still holds **except step 1(b)**, which drove a standalone `mfe:validate` that no longer exists here. Replace that step with the following; the rest is unchanged.

### 1(b), revised — the slot check is now a rule inside #309's command

```bash
./bin/run.js mfe:validate examples/meridian-station/meridian-console
```

Note the argument is now **positional**, not `--dir` — that's #309's signature. Expect six rules, the last one new:

```
  ✓ react-pinned
  ✓ manifest-package-sync
  ✓ shared-declared
  ✓ shared-version-sync
  ✓ runtime-declared
  ✓ slots-implemented
```

To make it fail, add `- id: ghost-rail` under `providesSlots` in that MFE's manifest and re-run. Expect exit 1, `MFE_INCONSISTENT`, and `✗ slots-implemented` naming `ghost-rail` — with the five dependency rules still passing beside it, which is the point of folding it in rather than bolting it on.

### New: the fleet-wide gate

This is the capability the standalone command never had — it wasn't CI-wired at all:

```bash
npm run check:mfe-consistency     # 21 MFE(s) checked — all internally consistent
```

`slots-implemented` evaluates only on the two slot providers (`abc-kids-home`, `meridian-console`) and is skipped on the other 19, since they declare no `providesSlots`.

### Build note

Unchanged and still required — `packages/runtime` and `packages/contracts` both moved, and `dist/runtime` is staged into MFEs and baked into the CLI image:

```bash
npm run build
```

### Merge order

**This PR is stacked on #309** and its base is `claude/issue-296-mfe-validate`. To test locally you get both branches' work; when #309 merges, GitHub retargets this to `main` and the diff narrows to the slot commits.

---

falese:
## Runbook fix — registry image was broken, pushed as `0cb1059`

@falese hit this running step 1(d):

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/slot-target.js'
  imported from /app/simple-registry.js
```

**My regression.** `1803b3c` extracted the slot matcher out of `simple-registry.js` into a sibling `slot-target.js`, but both registry Dockerfiles copy a single named file:

```dockerfile
COPY simple-registry.js ./     # slot-target.js never reaches the image
```

Fixed to `COPY *.js ./` — the build context holds exactly those two `.js` files and `.dockerignore` already excludes `node_modules`, so the glob is safe, and it's what stops the next extraction from silently shipping a broken image.

### Why the gates didn't catch it

CI never builds the control-plane images, and my own verification ran `node simple-registry.js` **from the source directory**, where `slot-target.js` sits right next to it. I verified the process, not the artifact — the one path this runbook actually tells you to take. That's the real miss, not the missing `COPY`.

### How it's verified now

No Docker daemon in my environment, so I reproduced each Dockerfile's `COPY` set into a clean directory — the same filesystem the image gets:

| | contents | result |
|---|---|---|
| before | `package*.json`, `simple-registry.js` | **`ERR_MODULE_NOT_FOUND`** — your exact error |
| after | `package*.json`, `simple-registry.js`, `slot-target.js` | boots: `🚀 Registry listening` |

Full runbook sequence against the fixed set:

```
register-station.sh   → 7 MFEs, 18 routes
GET /rules            → 7 mfes / 18 routes; meridian-console providesSlots: main, status, berth.{id}
POST meridian-console/mian    → HTTP 400  (names the declared alternatives)
POST not-yet/main             → HTTP 200  + ⚠️ warning
POST root                     → HTTP 200
```

### Runbook correction

Step 1(d) says `PORT=4500 node simple-registry.js`. That command is exactly what hid this bug — it runs from the source tree. If you're testing the containerised path, use:

```bash
docker build -t registry-test examples/meridian-station/control-plane/registry
docker run --rm -e PORT=4500 -p 4500:4500 registry-test
```

or just `docker compose up registry` from `examples/meridian-station`.

### Follow-up worth filing

**CI should build the two control-plane images.** A broken registry image currently fails for whoever runs the runbook next rather than in the build. The daemon Dockerfile has the same single-file `COPY` shape — not broken today since it has only one `.js` file, but the same trap. I left it alone rather than making a drive-by edit; say the word if you'd rather I harden it here.

Gates unaffected (Dockerfile-only change): lint 0 errors, typecheck clean, 21 MFEs drift-free, 21 internally consistent, both fleets' placement targets valid.

---

Issue #149, opened 2026-04-26
fix(dx): Rspack alias depth detection, broken dsl-mfe example, BFF generated without data: section


falese:
## Problem

Three concrete issues found during the ABC Kids build that erode trust in the tool and break the intended workflow:

### 1. Rspack alias path assumes exactly 2 levels of nesting

The generated `rspack.config.js` hardcodes:
```js
'@seans-mfe-tool/runtime': path.resolve(__dirname, '../../src/runtime/index.ts')
```

This only works if the project is at `examples/my-mfe/` (2 levels from repo root). At `examples/abc-kids/flappy/` (3 levels) it silently resolves to the wrong path. No build error at generation time — the error surfaces as a confusing `Cannot resolve` at build time.

**Fix**: Detect the depth at generation time and emit the correct relative path. Or better: use `require.resolve()` with a package name once `@seans-mfe-tool/runtime` is a proper published package.

### 2. The `dsl-mfe` example is broken for standalone development

`rspack.config.js` references:
- `entry: './src/index.tsx'` — file does not exist
- `template: path.join(__dirname, 'public/index.html')` — directory does not exist

Running `npm run start` in `examples/dsl-mfe/` fails immediately. For a tool that positions DX as a priority, the flagship example failing on first run is a significant trust issue.

**Fix**: Add `src/index.tsx` and `public/index.html` to the example. Alternatively, update the rspack config to match the actual file structure.

### 3. BFF scaffolding is generated even when the manifest has no `data:` section

For MFEs with no backend data layer, the generator still produces `server.ts`, `.meshrc.yaml`, BFF test stubs — files that will never be used. These create confusion about what's needed to run the MFE.

**Fix**: Skip all BFF output when `manifest.data` is absent. The condition already exists in the generator logic but isn't enforced consistently.

## Acceptance criteria

- [ ] `remote:generate` detects the output directory depth and emits the correct relative path for the runtime alias
- [ ] `examples/dsl-mfe/` has a working `src/index.tsx` and `public/index.html`
- [ ] `npm run start` in `examples/dsl-mfe/` boots without errors (with node_modules installed)
- [ ] No BFF files generated when manifest has no `data:` section
- [ ] All three fixes covered by regression tests

## Part of

Epic #139

falese:
[...]

Issue #146, opened 2026-04-26
feat(transparency): system:map — full MFE graph view, no running services required


falese:
## Problem

There's no way to see the whole system at a glance. When you have 5 MFEs and a shell, understanding what's registered, what's healthy, what capabilities each remote exposes, and how they connect requires opening multiple `mfe-manifest.yaml` files.

For a human taking over from an AI, or an AI picking up where a human left off, this missing system view is a significant friction point.

### `seans-mfe-tool system:map` command

```bash
seans-mfe-tool system:map           # text tree of the full system
seans-mfe-tool system:map --json    # machine-readable graph
seans-mfe-tool system:map --check   # include live health check results
```

### Human output

```
ABC Kids System
├── shell: abc-kids-shell  (port 3000)
│   ├── remote → abcKidsFlappy  http://localhost:3001/remoteEntry.js
│   └── remote → abcKidsHockey  http://localhost:3002/remoteEntry.js
│
├── remote: abc-kids-flappy  (port 3001)
│   ├── capability: PlayGame      shape: game-canvas   owner: developer
│   ├── capability: ShowCover     shape: ui-card       owner: developer
│   └── capability: GetGameInfo   shape: data          owner: developer
│
└── remote: abc-kids-hockey  (port 3002)
    ├── capability: PlayGame      shape: game-canvas   owner: developer
    ├── capability: ShowCover     shape: ui-card       owner: developer
    └── capability: GetGameInfo   shape: data          owner: developer

Generated: 2026-04-26  Branch: claude/plan-abc-kids-mfe-4v0OW
Shared singletons: react@18.2.0, @mui/material@5.14.0
```

With `--check` (makes live HTTP requests):
```
  ✓ abc-kids-flappy   http://localhost:3001/health  →  healthy  42ms
  ✗ abc-kids-hockey   http://localhost:3002/health  →  connection refused
```

### `--json` output

A graph structure:
```json
{
  "shell": { "name": "abc-kids-shell", "port": 3000, "remotes": ["abc-kids-flappy", "abc-kids-hockey"] },
  "remotes": [
    {
      "name": "abc-kids-flappy",
      "port": 3001,
      "mfName": "abc_kids_flappy",
      "capabilities": [
        { "name": "PlayGame", "shape": "game-canvas", "owner": "developer" }
      ]
    }
  ],
  "sharedSingletons": { "react": "^18.2.0" },
  "generatedAt": "2026-04-26T22:15:00Z"
}
```

[...]

Issue #145, opened 2026-04-26
feat(transparency): AI decision audit log + file ownership markers


falese:
## Problem

When an AI agent generates an MFE, it makes dozens of small decisions: which capabilities to expose, what shapes to use, what port to assign, what shared singleton versions to pin, whether to include a BFF. None of these decisions are recorded. A human reviewing the output sees the result but not the reasoning.

This creates a trust gap. The human head can't rationally audit what the AI head did because the decisions are invisible. Overriding becomes guesswork — you don't know what you'd be breaking.

### `.seans-mfe/audit.json` — the decision log

Every generation run appends a structured entry to `.seans-mfe/audit.json` in the project root:

```json
{
  "entries": [
    {
      "timestamp": "2026-04-26T22:15:00Z",
      "command": "remote:generate",
      "manifest": {
        "name": "abc-kids-flappy",
        "hash": "sha256:a3f9..."
      },
      "decisions": [
        {
          "id": "port-assignment",
          "value": 3001,
          "reason": "Specified in manifest endpoint field",
          "source": "manifest"
        },
        {
          "id": "mf-name",
          "value": "abc_kids_flappy",
          "reason": "Derived from manifest name by replacing hyphens with underscores (Module Federation identifier rule)",
          "source": "convention"
        },
        {
          "id": "shared-react-version",
          "value": "^18.2.0",
          "reason": "DEPENDENCY_VERSIONS constant in generator — pinned for singleton compatibility",
          "source": "generator-constant"
        },
        {
          "id": "bff-skipped",
          "value": false,
          "reason": "No data: section in manifest — BFF scaffolding omitted",
          "source": "manifest-absent"
        }
      ],
      "files": [
        { "path": "rspack.config.js", "action": "created", "owner": "platform" },
        { "path": "src/features/PlayGame/PlayGame.tsx", "action": "created", "owner": "developer" }
      ]
    }
  ]
}
```

### `decision` sources

| Source | Meaning |
|--------|---------|
| `manifest` | Value came directly from the manifest |
| `convention` | Derived by a documented rule (e.g. hyphen→underscore) |
| `generator-constant` | Hardcoded in the generator (pinned dependency version etc.) |
| `manifest-absent` | Decision made because a manifest field was absent |
| `default` | Generator default, manifest didn't specify |

### `seans-mfe-tool explain:decisions` command

```bash
seans-mfe-tool explain:decisions           # human-readable decision log
[...]

Issue #144, opened 2026-04-26
feat(shell): shell:init + shell:sync-types — generate the host app and keep remote types in sync


falese:
## Problem

There is no `shell:init` command. The ABC Kids host app was hand-written from scratch — about 8 files, ~350 lines, none of it generated. An AI agent building a system of MFEs has no pattern to follow for the host. Every time someone builds a new multi-MFE system, they reinvent the same Module Federation host wiring.

Additionally, the `remotes.d.ts` TypeScript declarations for federated modules must be manually maintained. When a remote adds or removes a capability, the shell's type declarations silently drift.

### `shell:init` command

```bash
seans-mfe-tool shell:init abc-kids \
  --remote flappy=http://localhost:3001 \
  --remote hockey=http://localhost:3002 \
  --port 3000
```

Generates a complete, working host app:

```
abc-kids-shell/
├── package.json          # rspack host, React 18, MUI — no CLI deps
├── rspack.config.js      # MF host with remotes[] auto-populated
├── tsconfig.json
├── public/index.html
└── src/
    ├── index.tsx          # ThemeProvider, createRoot
    ├── App.tsx            # placeholder layout with lazy remote loading
    ├── remotes.d.ts       # GENERATED — typed declarations for each remote
    └── components/
        └── RemoteLoader.tsx  # Suspense wrapper with error boundary
```

### `shell:sync-types` command

Reads each registered remote's manifest (via `/.well-known/mfe-manifest.yaml` or local path) and regenerates `remotes.d.ts`:

```bash
seans-mfe-tool shell:sync-types          # regenerate remotes.d.ts from live remotes
seans-mfe-tool shell:sync-types --json   # report what changed
```

Generated `remotes.d.ts`:
```ts
// AUTO-GENERATED by seans-mfe-tool shell:sync-types
// Do not edit by hand — run `seans-mfe-tool shell:sync-types` to update

declare module 'abcKidsFlappy/App' {
  const App: React.ComponentType<{ capability?: 'PlayGame' | 'ShowCover' | 'GetGameInfo' }>;
  export default App;
  export const PlayGame: React.ComponentType;
  export const ShowCover: React.ComponentType;
  export const GetGameInfo: React.ComponentType;
}
declare module 'abcKidsHockey/App' {
  // ... generated from hockey's manifest
}
```

The union type `'PlayGame' | 'ShowCover' | 'GetGameInfo'` is derived from the remote's manifest capabilities — not guessed.

[...]

PR #334, opened 2026-07-26
Re-derive epic #139 from measured friction, and close the agent-profile gaps it specified but never enforced

falese:
Asked whether epic #139 should be revisited "in light of what's been created." It should. Its framing became canon; its deliverables were overtaken. This PR reconciles the two and ships the first slice of what survives.

Governed by **ADR-077** (Accepted) · **ADR-078** (Proposed) · **PDR-008** (Accepted).
Full evidence: [`docs/platform-design-review/two-headed-giant-re-derivation.md`](https://github.com/falese/seans-mfe-tool/blob/claude/issue-139-strategy-review-nr90y1/docs/platform-design-review/two-headed-giant-re-derivation.md).

## Why not just execute the April spec

Epic #139 was written from one build (ABC Kids), before the platform had drift gates, a slot contract, a control plane in the reference apps, or a second reference app.

The decisive evidence is `examples/meridian-station/DX-REPORT.md` — a 22-item punch list written **live** while an agent built a 7-MFE reference app through the CLI *and* MCP. **None of its 22 items is on the #139 list.** What it actually found: the unpublished runtime (ADR-064/#252), React↔Angular template drift (#281), MCP `cwd` targeting (#279), and a dozen API-generator defects.

Three items were not merely late but wrong:

- **#145** (ownership markers + `audit.json`) — never built, and not coming. Ownership is the `overwrite` flag on the generation plan (`unified-generator.ts:800-802`) enforced by five regenerate-and-diff gates. ADR-074 settled the principle: *generating it removes the disagreement rather than checking for it.* A header comment is advisory; a gate is not.
- **#144** (`shell:init`) — its static `remotes[]` + `remotes.d.ts` spec is the opposite of ADR-055's daemon-driven shell. Meridian's shell ships `remotes: {}` on purpose.
- **#146** (`system:map`) — its discovery strategy parses that same empty `remotes` block, so it would report a fleet of zero remotes for the flagship reference app.

Meanwhile the contract ADR-033 *did* specify was unenforced: `--no-interactive` did not exist anywhere in `src/` or `packages/`, and typed exit codes applied only under `--json`.

### Governance

| File | What |
|---|---|
| `two-headed-giant-re-derivation.md` | Epic reconciled item by item against `main`; the DX-report evidence; the split |
| `ADR-077` | Replaces ADR-033's implementation table, not its framing. Ownership markers/audit log **withdrawn**; agent profile moves from `enforcement: convention` to `enforcement: code` |
| `ADR-078` + `PDR-008` | Control plane into the platform; `platform:init`; persistence as a manifest field; retire `@falese/daemon` |

Also corrects a defect in this review's own scorecard: it cited an ADR-033 `enforcer-config` block that **does not exist** (`enforcement: convention` is the whole of that frontmatter), inflating D3 and misdirecting D6. And it unblocks PDR-005, whose promotion was gated on PR #153 — closed, not merged, so the condition was unreachable.

### Code

- **`--no-interactive`** is now a `baseFlag` (`interactive`, `allowNo`), honoured independently of `--json`. Previously the only way to guarantee no prompt was to also give up human-readable output.
- **Typed exit codes on every failure path.** `BaseCommand` stamps `err.oclif.exit` and re-throws, so oclif still renders the message. Deliberately leaves alone errors that already carry a code (oclif usage errors) and errors that classify as `unknown` — relabelling a stray `Error` as a typed failure class is the opposite of what a branchable contract is for.
- **Fixed a latent double-envelope bug.** The success-path `process.exit` sat inside the `try`, so any throw from it fell into the `catch` and emitted a *second* envelope — breaking the one-line contract `BaseCommand` exists to keep. Five tests in `base-command.json-contract.test.ts` were red at baseline because of it.
- **Registry-driven conformance sweep.** Nothing iterated the command tree, which is why the above went unnoticed for three months. The stale catalog assertion (*"all 9 expected command schemas"* against 13 on disk) is replaced with a derived one.

## Two real defects the sweep found on its first run

**`schemas/deploy.json` advertised eight flags `deploy` does not have** — `registry`, `mode`, `namespace`, `domain`, `tag`, `memory`, `cpu`, `replicas`: a Kubernetes vocabulary the command never implemented (ADR-062 keeps production deployment out of it). Any agent passing one over MCP got a hard oclif parse failure. Fixed at source in `scripts/generate-schemas.ts`.

**A phantom `mfe:manifest.schema` MCP tool.** `src/mcp/sources/local.ts` recursed into `schemas/` and turned the DSL *document* schema (ADR-065) into a tool, carrying that schema's own description so it presented convincingly. Every call returned `command manifest.schema not found`. The source now requires a command-contract shape.

## Known gap, deliberately left open and pinned

There is no `mfe:build:*` tool. `build:check|dev|prod|docker` are real, working commands with no `schemas/*.json`, and `generate-schemas.ts` iterates a hand-written literal rather than the oclif registry — so a command with no hand-authored schema silently produces no tool. **An agent over MCP can scaffold and validate an MFE but cannot build it**, which is why the Meridian MCP lane scaffolded over MCP and then fell back to the CLI.

Not exempted and forgotten: an assertion fails if the uncovered set changes **in either direction**.

## Acceptance evidence

| Gate | Result |
|---|---|
| `npm run lint` | 0 errors, 0 warnings |
| `npm run typecheck` | clean |
| `npm test` | **2279 passed / 124 suites**, 0 failures |
| `npm run build` | OK |
| `npm run build:schemas` + diff | only the intended `deploy.json` correction |
| `npm run check:adr` | *"ADR library is consistent"* (78 ADRs, all 11 rules) |
| `npm run build:adr-index` + diff | regenerated and committed |

End-to-end, against the built CLI rather than mocks:

| Check | Result |
|---|---|
| `bff:init` in human mode | exit **64** (was 1) |
| `remote:init` in human mode | exit **65** (was 1) |
| `bff:init --json` | 1 stdout line, `ok:false`, `error.type: validation`, `error.code: 64` |
| `remote:init --no-interactive`, no TTY | fails instead of hanging |
| `mcp:serve` `tools/list` | **14 → 13**, phantom gone, all real tools intact |
| delete a `schemas/*.json` | sweep goes red (ratchet verified) |

## Issue surgery

**Closed:** #145 (withdrawn), #142 (superseded — features are developer-owned and hand-authored by design per PDR-001/#302; six stub variants would double the surface #281 exists to contain). #149 was already closed.

**Filed:** #329 `platform:init` epic · #330 one `describe()` engine (`explain`/`system:map`/`status`) · #331 CLI/MCP contract hygiene · #332 codegen packaging loose ends · #333 docs currency pass.

**Re-scoped:** #139 retitled *"Agent contract completion"*, original body preserved. #141/#143/#144/#146/#147/#148 annotated with what shipped, what moved, and why.

## Reviewer notes

- ADR-033 is **not** edited — CLAUDE.md forbids it. ADR-077 amends it from outside.
- Typed exit codes in human mode are an observable behaviour change: a command that exited 1 on a `ValidationError` now exits 64. That is what ADR-033 always specified, but any consumer branching on `=== 1` is affected.
- The 16 broken doc links reported by `docs:check-links` are all pre-existing and in files this PR does not touch (tracked in #333).

Refs #139, #329, #330, #331, #332, #333.

https://claude.ai/code/session_01EMRxKJjmJuCC3T8GVfHYWW

---
_Generated by [Claude Code](https://claude.ai/code/session_01EMRxKJjmJuCC3T8GVfHYWW)_


falese:
## Local-test runbook

Accurate to `69c2233`. Copy-pasteable end to end. No Docker, no running services needed.

### Setup

```bash
git fetch origin claude/issue-139-strategy-review-nr90y1
git checkout claude/issue-139-strategy-review-nr90y1
npm ci
npm run build      # required — the MCP + exit-code checks run the built CLI via bin/run.js
```

`npm run build` ends by regenerating the schema catalog. `git status` should stay clean afterwards; if `schemas/` is dirty, generation is non-deterministic and that is a bug.

### 1. `--no-interactive` no longer requires `--json`

```bash
mkdir -p /tmp/smt-check && cd /tmp/smt-check
node ~/seans-mfe-tool/bin/run.js remote:init "" --no-interactive < /dev/null; echo "exit=$?"
```

**Correct:** fails fast with a `BusinessError` and `exit=65`. Human-readable, coloured stderr — *not* a JSON envelope. Before this branch there was no such flag at all.

If it hangs waiting for input, `blockInteractivePrompts()` is not being reached from the non-JSON path — look at the `nonInteractive` branch in `BaseCommand.run()`.

### 2. Typed exit codes in human mode

```bash
cd /tmp/smt-check
node ~/seans-mfe-tool/bin/run.js bff:init > /dev/null 2>&1; echo "human exit=$?"   # expect 64
node ~/seans-mfe-tool/bin/run.js remote:init "" > /dev/null 2>&1; echo "human exit=$?" # expect 65
```

**Correct:** `64` (validation) and `65` (business). On `main` both are `1`.

First place to look if wrong: `withTypedExitCode()` in `packages/oclif-base/src/BaseCommand.ts`. It intentionally skips errors that already carry `oclif.exit` and errors classifying as `unknown`, so a raw `Error` still exits 1 — that is by design, not a miss.

### 3. `--json` still emits exactly one envelope

```bash
cd /tmp/smt-check
node ~/seans-mfe-tool/bin/run.js bff:init --json 2>/dev/null > out.json; echo "exit=$?"
wc -l < out.json          # expect exactly 1
python3 -m json.tool out.json | head -20
```

**Correct:** `exit=64`, **one** line, `"ok": false`, `error.type: "validation"`, `error.code: 64`. The line count matters — the double-envelope bug this PR fixes showed up as `2`.

### 4. The MCP tool catalog

```bash
cat > /tmp/mcp-list.mjs <<'EOF'
import { spawn } from 'node:child_process';
const proc = spawn('node', [process.env.HOME + '/seans-mfe-tool/bin/run.js', 'mcp:serve'],
                   { stdio: ['pipe','pipe','inherit'] });
let buf = ''; const pending = new Map(); let id = 1;
proc.stdout.on('data', c => { buf += c; let n;
  while ((n = buf.indexOf('\n')) >= 0) { const l = buf.slice(0,n).trim(); buf = buf.slice(n+1);
    if (!l) continue; let m; try { m = JSON.parse(l); } catch { continue; }
    if (m.id && pending.has(m.id)) { pending.get(m.id)(m); pending.delete(m.id); } } });
const send = (method, params) => new Promise(res => {
  const i = id++; pending.set(i, res);
  proc.stdin.write(JSON.stringify({ jsonrpc:'2.0', id:i, method, params }) + '\n'); });
await send('initialize', { protocolVersion:'2024-11-05', capabilities:{},
                           clientInfo:{ name:'check', version:'0' } });
proc.stdin.write(JSON.stringify({ jsonrpc:'2.0', method:'notifications/initialized' }) + '\n');
const r = await send('tools/list', {});
console.log('TOTAL:', r.result.tools.length);
for (const t of r.result.tools) console.log(' ', t.name);
proc.kill(); process.exit(0);
EOF
node /tmp/mcp-list.mjs
```

**Correct:** `TOTAL: 13`, and **no `mfe:manifest.schema`**. On `main` it is 14 and the phantom is there; calling it returns `command manifest.schema not found`.

Also confirm what is *still* missing — there is no `mfe:build:*` in the list. That is the known gap called out in the PR body, not a regression.

### 5. The conformance ratchet actually ratchets

```bash
cd ~/seans-mfe-tool
mv schemas/slots-validate.json /tmp/
npx jest src/oclif/__tests__/command-conformance.test.ts 2>&1 | grep -E "✕|Tests:"
mv /tmp/slots-validate.json schemas/
```

**Correct:** two failures — `slots:validate has an MCP schema...` and `the build surface is the only uncovered group`. Then green again once restored. If it stays green with the file removed, the sweep is not actually reading the catalog.

### 6. Full gates

```bash
npm run lint          # 0 errors, 0 warnings
npm run typecheck     # clean
npm test              # 2279 passed / 124 suites
npm run check:adr     # "ADR library is consistent" (78 ADRs)
npm run build:schemas && git diff --exit-code schemas/     # clean
npm run build:adr-index && git diff --exit-code docs/spec.md docs/architecture-decisions/README.md
```

### Known non-issues

- `node scripts/check-doc-links.js` exits 1 with 16 broken links. All pre-existing, none in files this PR touches — tracked in #333.
- `deploy --json` no longer advertises `registry`/`namespace`/`replicas` etc. That is the fix, not a regression: the command never had those flags.

---

falese:
## Update — the build surface is in, via the registry rather than four new files

Pushed `998c975`. The ask was to add the build surface *and* make any new oclif command automatically appear as an MCP tool. The second half is what makes the first half stay true.

### The literal had drifted in every direction available to it

Adding four schema files would have fixed one symptom and left the mechanism intact. Deriving from the registry killed four defects at once:

| Defect | Consequence |
|---|---|
| `build:check\|dev\|prod\|docker` had no entry | No `mfe:build:*` tool — an agent could scaffold and validate an MFE, then not build it |
| `deploy` advertised 8 flags it doesn't have | Hard oclif parse failure for any agent that used one |
| **`remote:init` never advertised `framework`** | **No agent could scaffold an Angular MFE over MCP at all.** Both MCP-scaffolded Meridian MFEs are React — that wasn't a coincidence |
| `buildArgv` hardcoded `['name','spec']` as positionals | `spec` is a flag on every command that has one, so `mfe:api` with a spec produced a second positional a `strict` command rejects |

### How it works now

`scripts/generate-schemas.ts` loads the live registry via `Config.load()` — the same resolution the CLI does at startup, first-party plugins included — and derives every input schema from the command's own flag and arg declarations. Declared args are carried through as `x-positional`, so `buildArgv` reads them instead of guessing.

A new command is catalogued automatically. What it *cannot* do is ship silently: output schemas can't be introspected from a TypeScript return type, so a catalogued command without one is a **build failure** with a message naming both options — author the output contract, or exclude it in `CATALOG_EXCLUDED` with a reason.

```
$ npm run build:schemas
...
17 schema(s) generated in schemas/
3 command(s) excluded by policy: mcp:serve, schemas, remote:init-angular
```

The three exclusions are things that aren't capabilities (the server itself, the catalog lister, a deprecated alias) — not parked gaps. Orphans are removed too: a schema whose command no longer exists advertises a tool that can't run, which is the phantom `mfe:manifest.schema` failure from the other side.

### Verified against the running server

```
TOTAL TOOLS: 17
  mfe:build:check   mfe:build:dev   mfe:build:docker   mfe:build:prod   ...
```

And an actual round trip — `mfe:build:check` against `examples/abc-kids/flappy`:

```json
{ "plugin": "react-rspack", "framework": "react", "bundler": "rspack",
  "checks": [ { "tool": "node", "required": ">=18", "found": "22.22.2", "ok": true },
              { "tool": "rspack", "required": ">=0.5", "found": null, "ok": false,
                "fix": "npm i -g @rspack/cli" } ],
  "allPassed": false }
```

Structured, typed, matching the published output schema.

### Gates

lint 0/0 · typecheck clean · **2309 tests / 126 suites** (was 2279/124) · build OK · `build:schemas --check` clean · `check:adr` consistent.

ADR-077 gains **§5** recording the decision and its boundary — the derivation covers the *input* half; output schemas are still hand-authored and unverified against actual return values, which needs type-level extraction this doesn't attempt.

### Runbook delta

Step 4 of the runbook above now expects **17** tools, not 13. To exercise the new surface:

```bash
node /tmp/mcp-list.mjs | grep build     # expect four mfe:build:* entries
```

To confirm the ratchet, add a command with no output schema (or delete an `OUTPUT_SCHEMAS` entry) and run `npm run build:schemas` — it should fail with the message naming both options rather than emitting `{}`.

---

falese:
## Update — output schemas derived too; two more commands were publishing a broken contract

`32f47d6` (the targeted fix) and `e55fce8` (the generator-side fix).

### What the audit found

You were right that the result types are defined — all 20 commands declare one. I resolved each through the TypeScript compiler and diffed against the published schema. **Two disagreed**, both `additionalProperties: false`:

| Command | Missing from published schema | Effect |
|---|---|---|
| `remote:generate` | `preserved` | Validating agent rejected **every** response — and couldn't see the field reporting whether its own hand-written feature code survived regeneration |
| `bff:validate` | `manifest`, `meshConfig` (both non-optional) | Same |

There was a second-order finding: `src/oclif/__tests__/json-contract.test.ts` opens with *"validates every envelope against the CommandResult meta-schema **plus the per-command output schema from `schemas/`**."* It doesn't — `ajv.compile(ENVELOPE_SCHEMA)` is the only compile. The output half was never implemented, which is why both drifted silently.

**How they were found is the argument for deriving.** My hand audit over `src/oclif/results.ts` caught the first and missed the second, because `BffValidateResult` lives in `packages/bff-plugin/src/types.ts`. The generator found both.

### Why derive rather than validate at runtime

The chain is: implementation → declared `T` → published schema. TypeScript **already** enforces the first link — `BaseCommand<T>` with `runCommand(): Promise<T>` is compiler-checked. Only `T` → schema was unverified.

So deriving closes the chain with **no runtime validation anywhere**. I deliberately did *not* add output validation to `BaseCommand`: it would re-check a link the compiler already guarantees, at per-call cost. The alternative — implementing the check that docstring promises — would need a success-path fixture per command (a scaffolded MFE for `remote:generate`, a Docker daemon for `build:docker`). Expensive and flaky, for strictly less coverage.

### Scope of the converter

`src/oclif/type-to-schema.ts` handles what command results actually use: primitives, arrays, string-literal unions → enum, nullable primitives, nested interfaces, inline object literals, flattened inheritance (`MutatingResult` spreads rather than `$ref`s), and JSDoc → `description`. Anything it can't represent faithfully degrades to an open schema rather than a confident wrong one — a result type too complex to express is a signal about the result type.

Union type arrays are sorted, because TS doesn't guarantee member order and these are under a byte-for-byte drift gate. Verified: two consecutive generations are byte-identical.

`generate-schemas.ts` went **493 → 211 lines** and now contains no command knowledge at all.

### One correction to my own framing

I initially flagged that `bff-validate.json` grew 1.3 KB → 22.5 KB (`manifest: MFEManifest` expands in full) as a token cost for agents. **That was wrong.** `tools/list` carries `name`, `description`, and `inputSchema` only — the output schema never reaches the agent. Measured: **11.7 KB, ≈2.9k tokens** for the whole 17-tool catalog. Output schemas serve `schemas/`, the `schemas` command, and response-validating clients.

Related follow-up, not done here: MCP does support an optional `outputSchema` on tool definitions. Now that ours are trustworthy, exposing them is a real option — but it'd add meaningfully to catalog size, so it deserves its own decision.

### Gates

lint 0/0 · typecheck clean · **2323 tests / 127 suites** · build OK · `build:schemas --check` clean · generation deterministic · `check:adr` consistent.

ADR-077 §5 now covers both halves and records the boundary: no runtime validation, and a narrow converter by design.

---

falese:
## Update — build parsers (`c04113c`) and response validation (`e6dcd04`)

### The build error was worse than I described

I said earlier that agents got "a wall of raw stderr." Wrong. Both plugins did:

```ts
const stderr = (err as { stderr?: string }).stderr ?? '';
return { errors: [{ message: stderr, category: 'unknown' }] };
```

and **rspack and tsc both write diagnostics to stdout.** Verified against the real `execSync` failure path — `err.stderr` is `''`, everything is in `err.stdout`. So every failing build reported exactly one error whose message was **the empty string**. An agent learned that something broke and nothing else.

### Parsers written against real output, not documented formats

Fixtures in `packages/contracts/src/__tests__/fixtures/` are captured by breaking a project and running the actual tools. That surfaced details the docs omit:

- rspack positions live on the `ERROR in <file> <line>:<col>` header, not beside the message
- a **parse-error** header has *no* position — the only line number is the gutter of the source excerpt
- rspack nests causes (`× Module parse failed:` → `╰─▶ × JavaScript parse error: ...`); the innermost names the real problem
- the message contains an apostrophe (`Can't resolve`), so extracting the specifier by first-quote match yields `"t resolve "`

Three rules: classify by **actionable cause** (`TS2307` is a type diagnostic but the fix is a missing dependency → `dependency`); distinguish a relative specifier (path mistake) from a bare one (missing install); and **never go quiet** — unrecognised output becomes one `unknown` error carrying the text, and a failure that wrote nothing falls back to the thrown error's message.

End to end, a real failing rspack build through `build:prod --json`:

```json
{ "file": "./src/index.js", "line": 1, "column": 1, "category": "dependency",
  "message": "Module not found: Can't resolve './does-not-exist' in '…/src'",
  "suggestion": "Check the path — no module at './does-not-exist' relative to the importing file." }
{ "file": "./src/index.js", "line": 2, "column": 1, "category": "dependency",
  "message": "Module not found: Can't resolve 'react' in '…/src'",
  "suggestion": "Install the package: npm install react" }
```

Two errors, correct positions, **correctly different** suggestions. Exit 65.

Also fixed a restatement the derivation surfaced: `BuildProdResult` declared its own inline copy of `BuildError`'s fields and had fallen behind — no `code`, and `category: string` instead of the union, so the schema advertised a free-form string where the contract has six known values.

---

### Response validation caught a defect in the derivation itself

This is the part worth reading. I argued in the last update that derivation made runtime validation redundant, because TypeScript already enforces implementation → declared `T`.

**That holds only where the compiler is actually checking, and here it wasn't.** The repo compiles with `strict: false`, so `strictNullChecks` is off and TypeScript collapses `string | null` to `string`. The derived schema asserted `EnvCheckResult.found` was always a string — while `build:check` really returns `null` for a tool it can't find:

```
build:check returned data its published schema rejects:
data/checks/2/found must be string
```

Nothing but running the command and validating the result would have found that. The generator now reads types with `strictNullChecks: true` regardless of repo settings — `| null` was authored deliberately and describes what the command does.

So I've corrected ADR-077: derivation keeps the schema honest about the **declared type**; response conformance keeps the declared type honest about **runtime**. Complements, not substitutes.

Seven commands have a success-path invocation cheap enough for the suite (`adr:status`, `adr:validate`, `build:check`, `mfe:validate`, `slots:validate`, `deploy --dry-run`, `remote:init --dry-run`). The other ten name their infrastructure cost — needs a Docker daemon, starts a long-lived server, needs `node_modules` in the target — and a ratchet asserts every catalogued command is in one list or the other.

Still deliberately absent: runtime validation inside `BaseCommand`. That puts cost on every production call to catch what one test run catches.

### Gates

lint 0/0 · typecheck clean · **2362 tests / 130 suites** · build OK · `build:schemas --check` clean · `check:adr` consistent.

ADR-077 gains §6 (build classification) and a recorded correction in Boundaries.

---

falese:
## Update — strict TypeScript (`d64d9f5`) and real Angular output (`ba2eee9`)

### On Bun vs tsc

Bun doesn't remove the need for tsc, and the `strict` question is unrelated to it. Demonstrated rather than asserted — a file with two type errors:

```
=== bun ===   bun ran this anyway: 1 undefined     bun exit=0
=== tsc ===   error TS2322 / error TS2339          tsc exit=2
```

Bun transpiles; it never typechecks. tsc is required for three separate jobs here: `npm run typecheck`, emitting `dist/` for the published Node entry (ADR-020 keeps Bun dev-only), and reading types via the compiler API for schema derivation. **Bun ignores `strict` entirely**, so enabling it affects only tsc runs.

### strict: false → true, repo-wide

Cost was far smaller than I expected: **10 errors, all in one file**, plus 6 in `packages/runtime` (own build config). Every tsconfig set `strict: false` explicitly and the package ones don't extend the root, so it was 8 files.

The fixes were real bugs, not type noise:

| Fix | What it was |
|---|---|
| `unified-generator` render-model arrays | bare `[]` inferred as `never[]`; now named types shared by declaration and assignment so they can't drift |
| `retry-wrapper` | `lastError` read for telemetry on the first retry *before* any assignment |
| `remote-mfe` | `root` never narrowed out of `Map.get()`'s `\| undefined` because `createRoot` arrives through a `@ts-ignore`'d import |
| `layout-adaptors` | `remoteEntryUrl`/`scope`/`module` are optional on the wire and were passed through undefined — surfacing as `script.src = "undefined"`, a 404 on a nonsense URL. Now a clear error naming which fields the placement rule lacks |

**One defect found and deliberately not fixed.** The DI branch of `executeLifecycle` passes a whole hook *entry* (`{name: config}`) to `LifecycleExecutor.execute`, whose signature says `LifecycleHook` — while the default branch expands each entry and calls `executeHook` per hook. The two paths disagree, and nothing caught it because the DI branch was untypechecked and its one test asserts `expect.any(Object)`. I typed the interface to match actual behaviour and flagged it in the doc comment: changing what a DI extension point receives is a lifecycle-semantics decision, not a strict-mode cleanup. Worth its own issue.

Schemas are byte-identical after the change — expected, since the generator already pinned `strictNullChecks` when reading types. That pin stays so the contract holds regardless of future tsconfig edits.

---

### Angular: the hand-written cases were wrong

You were right to want this tested. I installed the real toolchain and broke `examples/abc-kids/multiplication-quiz`. Two findings:

**1. `ng` writes to stderr — the opposite of rspack and tsc.** So the two framework paths were failing in two *different* ways, and my earlier summary was only half right:

| Toolchain | Writes to | What the agent got |
|---|---|---|
| rspack, tsc (React) | **stdout** | `message: ''` — empty stderr |
| `ng build` (Angular) | **stderr** | the whole raw blob as one `unknown` error |

I'd said every failing build reported an empty message. True for React, not Angular. ADR-077 §6 now states both.

**2. `ng` colours its output even when piped.** ANSI escapes sit between the file, line, column and code — so every pattern failed. `grep 'error TS'` finds **zero** matches in the captured fixture for exactly that reason. And its module errors use a webpack single-line form (`./file:line:col-col - Error: Module not found: Error: …`) sharing no shape with rspack's `ERROR in` header, with a doubled `Error:` prefix.

Against that fixture the parser produced **1 unrecognised blob** before, and **11 classified errors** after — verified end to end through `AngularWebpackPlugin.buildProduction` against the actual broken project:

```json
{"file":"src/app/app.component.ts","line":22,"column":9,"code":"TS2322","category":"type",
 "message":"Type 'number' is not assignable to type 'string'."}
```

### Gates

lint 0/0 · typecheck clean · **2367 tests / 130 suites** · build OK (full strict) · `build:schemas --check` clean · `check:adr` consistent.

---

falese:
## Update — `lifecycleExecutor` deleted (`0761de6`, ADR-079)

### Blast radius, measured

Base-class changes usually deserve the worry, so I measured instead of assuming:

```
packages/runtime/src/base-mfe.ts   4 references
  62:  export interface LifecycleExecutor {
  91:  lifecycleExecutor?: LifecycleExecutor;
  471: if (this.deps?.lifecycleExecutor) {
  473:   await ...execute(hookEntry, ...)
```

**This is not a regenerate scenario.** No template references it, so there is no generated output to change:

```
check:mfe-drift:check     → 21 MFE(s) checked — no generator-owned drift
check:mfe-consistency     → 21 MFE(s) checked — all internally consistent
git status examples/ templates/  → (empty)
```

Identical before and after. And it was never in `packages/runtime/src/index.ts`, so no consumer could even name the type.

### The finding that decided it

**The seam it was reaching for already exists, and it's `customHandlers`.**

```ts
// every generated mfe.ts, both frameworks:
super(manifest, { customHandlers: handlerRegistry });

// base-mfe.ts invokeHandler:
const injected = this.deps?.customHandlers?.[handlerName];
```

Handler-level DI is shipped, wired into both template trees, and load-bearing for ADR-040. Because it substitutes at the *innermost* call, ADR-002's guarantees — handler arrays, `contained`, `mandatory`, main-phase propagation, telemetry — still apply with no cooperation from the substitute.

`lifecycleExecutor` sat *around* the phase loop and reached none of that. It was a second, broken, unused version of a seam that already worked — one positioned outside the contract instead of inside it.

Three compounding problems, all now moot: it bypassed ADR-002 entirely; its declared type disagreed with what it was passed (found only by enabling strict); and being called once per entry, sequentially, awaited, it could never see a phase whole — so it couldn't have delivered the ADR-028 parallelism it would most plausibly serve.

### The tests now assert something

They previously used `expect.any(Object)` and `any`, which is why the type mismatch survived. Replaced with:

- every hook entry in a phase runs **in declared order** through `customHandlers`
- `contained: true` still contains a failure **and lets the following hook run** when execution is substituted — exactly what the deleted branch skipped

### Gates

lint 0/0 · typecheck clean · **2368 tests / 130 suites** · build OK · drift + consistency gates unchanged at 21 MFEs · schemas clean · `check:adr` consistent.

ADR-079 records the decision: one seam for substituting execution, and it sits inside the lifecycle contract. A phase-level seam, if ADR-028 ever needs one, should arrive with that ADR rather than be kept in reserve against a guess.

---

falese:
## CI failure diagnosed and fixed (`b7f5776`) — and it wasn't the test

All 14 failures in `output-schema-conformance.test.ts` had one cause: **CI never built the CLI.**

The `test` job compiled only `packages/contracts` and `packages/oclif-base`, leaving `dist/`, `oclif.manifest.json`, and five workspace packages absent. `node bin/run.js` then either resolved zero commands —

```
"why": "command adr:status not found"
"why": "command remote:generate not found"
```

— or crashed inside oclif's `findCommand` with `MODULE_NOT_FOUND`, giving empty stdout and the `SyntaxError: Unexpected end of JSON input` on the other nine.

Reproduced locally by moving the same dists aside. In that state even `schemas --json` returns **0 bytes, exit 1**.

### The part worth fixing properly

I first assumed my test was the odd one out, since the pre-existing spawn-based test passed. It didn't pass — **it was excluded from CI**:

```yaml
--testPathIgnorePatterns="json-contract.test.ts"
--testPathIgnorePatterns="remote-mfe.integration.test.ts"
```

So the CLI's JSON-envelope contract — what ADR-018 exists to guarantee, and what every agent integration depends on — has had **no CI coverage at all**. The workaround for an unbuilt CLI was to hide the test that would have reported it.

That's the same failure mode as the `doc-links` job earlier in this PR: a check that can't fail isn't a check. It just takes longer to notice.

### Fix

- `Build the CLI` now runs `npm run build` — the same command the local-test runbook gives a human, which is the standard CI should be held to
- both exclusions removed; each suite passes against a built CLI (15 and 18 tests)

Verified locally with CI's exact invocation: **130 suites, 2377 tests, 0 failures.**

### Runbook delta

Step 4 of the runbook now expects **17** MCP tools, not 13. Otherwise unchanged — and worth noting the runbook always said `npm ci && npm run build`, which is precisely what CI wasn't doing.

I'll keep watching this PR until it's merged or closed.

---

PR #335, opened 2026-07-26
refactor: single-source the platform contract in @seans-mfe/contracts

falese:
The ten platform capabilities and the six-state MFE lifecycle machine had no
definition in code — they were re-declared at ten sites across the DSL, the
runtime, codegen, and the EJS templates. Five of those sites listed nine
capabilities: `updateControlPlaneState` was added to the runtime and to the two
`mfe.ts.ejs` templates and never to the rest.

That was a live defect, not just duplication. Codegen classifies a manifest
capability as platform by looking it up in its own nine-entry map; a miss means
domain. So a manifest declaring `UpdateControlPlaneState` generated a feature
stub, a smoke test calling it as user code, and an `import type
{ UpdateControlPlaneStateOutputs }` that no emitted file defines.

Add `packages/contracts/src/platform-contract.ts` as the single source:
capability names, both manifest spellings, `do*()` wrapper methods, result
types, HTTP surface, and each capability's pre-states and enter/exit/error
transitions, alongside the lifecycle states and their transition table. Pure
frozen data plus four predicates — contracts stays zero-dependency (ADR-061),
which is what lets zod-bound DSL, the staged runtime, and build-time codegen all
read it.

Every definition site now consumes it. `MFEState` and `VALID_TRANSITIONS` keep
their runtime names — generated MFEs import them — as aliases of the canonical
type and table. EJS cannot import, so the render model carries
`platformCapabilityNames` and the four templates filter against that.

Drift is pinned by behavior, not by comparing literals: the runtime pin drives
every capability from every state and holds the guards to the contract's
pre-states; the codegen pin asserts no platform capability is ever generated as
domain, in both framework variants. Both fail against the pre-change code (5 of
7 codegen cases).

Generated output is byte-identical — the template arrays lived inside `<% %>`
blocks — so no example MFE needed regenerating.

ADR-080


falese (commented):
Are we so sure that the contracts package was designed as primitives I'm pretty sure that it's always been intended as the platform "spec"

falese:
You're right, and I've corrected the ADR. Pushed in 4fbc1a1.

I went and checked rather than just conceding, and the evidence is on your side:

- **ADR-054** is titled *"Control-Plane Message Protocol **as a Shared Contract** in `@seans-mfe/contracts`"* and its Decision §1 reads "`@seans-mfe/contracts/messages` is **the single source of truth** for the control-plane wire protocol." That's a spec, not a primitive.
- **ADR-056** puts the MFE presentation boundary there. **ADR-036** puts the framework-plugin base there. **ADR-069/073** put the slot grammar and slot-contract logic there.
- The package's own description leads with "Shared runtime **contracts**."

So the "started as primitives, now drifting" framing in my Consequences section was just wrong — the remit was established well before this PR, and the capability set plus lifecycle machine are about the most contract-shaped things in the repo. Nothing here expands what ADR-054 already set.

I replaced that bullet with a **"Not a cost: where it lives"** paragraph making the ADR-054 lineage explicit, and kept the one cost that *is* real: the coupling's blast radius. Changing `PLATFORM_CAPABILITY_SPECS` now moves DSL validation, codegen output, and runtime guards together — a mistake becomes uniform rather than partial, and the pin tests are what stand between a typo and every generated MFE.

**Separately — the `quality` CI failure was mine, and it also dented a claim in the ADR.** The EJS comment block I added above the domain-capability filter in both `mfe.test.ts.ejs` templates emitted its trailing newline into every generated file, drifting all 21 committed examples. I'd reported the output as byte-identical on the strength of `check:mfe-consistency`, which checks manifest/package agreement, not generated bytes — `check:mfe-drift` is the gate that actually proves that, and I hadn't run it. Comment is now folded into the scriptlet; `check:mfe-drift` reports no drift, and the Boundaries bullet records the trap and names the right gate.

---

PR #338, opened 2026-07-27
Clear the issues dangling off the ADR register: #199, #247, #318, #320, and an observability schema

falese:
> **Retitled.** This PR was opened on its first commit and accumulated seven more. The description below covers all eight. The template typed-errors work that was briefly here has moved to **#342**, stacked on this branch.

## Why

The ADR register (ADR-075) links decisions to work two ways — `impl.refs` for "what must happen before this is finished", `tracked-by` for "what is currently open about it". Neither is checked against GitHub, because CI must not put a network call in the correctness path, so both rot. The register named 24 issues; 5 had already closed, and one ADR was finished without anyone flipping its status.

This works that backlog: verify what is done, close it, and implement what is not.

`adr:status --outstanding` goes from **9 decisions to 7**, and every remaining entry names an issue that is genuinely open.

## What's in it

**`3b36857` — register reconciliation.** Pruned 5 closed refs (#241, #264, #281, #317, #321) from `tracked-by`. ADR-065 `Accepted, phased` → `Implemented`: its tracker closed when phase 2 landed, `docs/api/` is seeded and the drift gate armed, so the decision is carried by code rather than a plan.

**`09fe8e7` — #199, demo mode on the standalone BFF path.** The issue located the gap in `bff:init`; it was actually in `extractMeshConfig`, which yaml-dumps the manifest's `data:` block and dropped `mockSwitch` entirely — the standalone path never renders `meshrc.yaml.ejs` at all. `applyMockSwitch` now expands the flag into the `resolversComposition` over `Query.*` and guards `responseCache` so live and mock are never cross-served; `ensureMockSwitchFiles` writes the composer before `mesh build`, when Mesh resolves the reference.

**`82e8855` — #318, ADR-026's load criteria.** Eight of eleven now asserted against the **real** `doLoad`. They could not have been verified before: `MFETestHarness` replaces `doLoad` wholesale with a fixture, so every existing load test asserted against a mock. That surfaced a defect the mock could not — every failure reported `phase: 'entry'` whatever actually failed, sending whoever read the telemetry to the network when the fault was the shared scope or the manifest. Three criteria remain, all blocked on one missing decision (see below).

**`49455dc` — #247, render fallback.** `createErrorBoundary` takes an MFE-supplied fallback resolved from the container's `./ErrorBoundary` expose, and emits `render-fallback-applied` — the mount succeeds when the boundary catches, so nothing previously recorded that a user was looking at a fallback. Angular had no fallback at all: a throw out of `createApplication` or bootstrap left the host element empty, the white screen ADR-044 exists to prevent.

**`06c8213` — ADR-081, one observability schema.** #322 filed ADR-039's logger as missing; the rationale under it had expired, since `redirectStdoutToStderr()` has owned the `--json` stdout contract since ADR-018 shipped. The real gap was that five surfaces describe themselves and none can be joined — the runtime emits span-shaped data with no span type, the BFF already emits real OpenTelemetry, and `correlationId` is a trace id doing three unrelated jobs. `PlatformEvent` is OTel-shaped with W3C `traceparent` propagation, and **no `@opentelemetry/*` dependency anywhere** — `packages/runtime` is federated into a browser bundle and the ADR-056 boundary test forbids it.

**`a4e1411` — a live bug plus the command's own span.** `postrun` hooks never ran under `--json`, because `run()` exits once the envelope is written and oclif fires them after the command returns. Measured: human mode ran them, `--json` did not — silently disabling them on the MCP path (ADR-019 spawns each tool call as its own `--json` invocation) and in CI. Every invocation now also emits a `cli.command` span on the envelope's trace, in both modes.

**`8c2fc11` — #320, typed errors.** All 39 raw throws converted. ADR-017 → `Implemented`, with a guard that reports file and line and pins its own matcher.

## ⚠️ Behaviour change worth knowing

**Exit codes change on the paths #320 converted.** A command that failed with `unknown` → 70 now exits 64 (validation), 66 (network), 69 (system) or 77 (security). That is the point of ADR-017 + ADR-018 — it makes failures branchable — but anyone scripting around exit 70 will see it.

Two mechanisms read `error.type`: `formatError` classifies by it, and `endSpan` attributes from it. Before this, a missing flag and a genuine crash were the same thing to both.

## Still blocked

**#318 cannot close** without deciding where retry configuration lives. `withRetry` exists and is tested (ADR-030), `handleError` states plainly that retry is that wrapper's job — but nothing calls it, and the config ADR-026 assumes (`errorHandling.retry.maxAttempts`) has no field in `packages/dsl/src/schema.ts`. Three criteria depend on it. It needs its own ADR rather than an implementation that invents the field. Recorded in ADR-026 and on the issue.

## Verification

- 139 suites, **2534 tests**, lint, typecheck, `build`, `test:ci`
- `check:adr`, `build:adr-index`, `check:mfe-drift:check`, `check:mfe-consistency` all clean
- `docs/api` regenerated (contracts and oclif-base are public surfaces); no schema drift
- Trace propagation verified end to end: an inherited `TRACEPARENT` is adopted, a malformed one starts a fresh trace instead of failing the command

Closes #199, closes #247, closes #320
Refs #318, #322, ADR-017, ADR-018, ADR-026, ADR-044, ADR-056, ADR-075, ADR-081


falese:
## Local-test runbook — `8c2fc11`

Every output below was captured by checking out this SHA and running the commands, not from the tip of the stack.

```bash
git fetch origin && git checkout claude/adr-dangling-issues-ir9eg8
npm ci
npm run build          # required — dist/ is what bin/run.js executes
```

`npm run build` is not optional here. This PR changes `packages/runtime` and `packages/oclif-base`, and **`npm run typecheck` does not compile `packages/runtime`** — build is the gate that does (#341).

### 1. The exit-code change — the one thing that could bite a caller

```bash
node bin/run.js mfe:validate /nonexistent-xyz --json; echo "EXIT=$?"
```

```json
{"ok":false,"error":{"type":"validation","code":64,"message":"Invalid or missing mfe-manifest.yaml in /nonexistent-xyz: …","retryable":false,"userFacing":true,"details":{"field":"manifest","constraint":"invalid"}},"warnings":[],"telemetry":{"durationMs":69,"correlationId":"cbc9cd4c-…","traceId":"45c83aa0fddc03e096ceaa519a0850a9"}}
EXIT=64
```

**`EXIT=64`, not 70.** Before this PR that path threw a raw `Error` and exited 70 (`unknown`). Anything scripting around exit 70 sees this. The envelope now carries `traceId` alongside `correlationId` — those were one field doing three jobs.

Wrong result: `EXIT=70` and `"type":"unknown"` means you're on a stale `dist/`. Re-run `npm run build`.

### 2. Trace context is adopted, not replaced

```bash
TRACEPARENT="00-11111111111111111111111111111111-2222222222222222-01" \
  node bin/run.js mfe:validate examples/abc-kids/flappy 2>&1 | grep cli.command
```

```json
{"traceId":"11111111111111111111111111111111","spanId":"6ccc8b74dfee019a","parentSpanId":"4868025e9614e933","name":"cli.command","startTime":"…","status":"ok","attributes":{"service.name":"seans-mfe-tool","cli.command":"mfe:validate"},"duration":20}
```

The `traceId` is **the one you supplied** — a CI job or an MCP parent can join CLI invocations into one trace. Now feed it garbage:

```bash
TRACEPARENT="garbage" node bin/run.js mfe:validate examples/abc-kids/flappy 2>&1 | grep cli.command; echo "EXIT=$?"
```

A fresh random `traceId` and `EXIT=0`. A malformed header from an upstream system starts a new trace rather than failing the command — telemetry must not be able to break the CLI.

Note the span goes to **stderr**, and appears under `--json` too, where stdout carries exactly one envelope line (ADR-018). Worth confirming: `node bin/run.js mfe:validate examples/abc-kids/flappy --json 2>/dev/null | wc -l` → `1`.

### 3. postrun hooks under `--json` — the live bug

This is the one that was silently broken: `run()` exits once the envelope is written, and oclif fires `postrun` after the command returns, so under `--json` they never ran at all. That is the MCP path (ADR-019 spawns every tool call as its own `--json` invocation) and CI.

Behavioural, so it is pinned by test rather than eyeballed:

```bash
npx jest packages/oclif-base/src/__tests__/base-command.observability.test.ts
```

```
✓ runs postrun hooks before exiting
✓ runs them on the failure path too
✓ passes the result through so a hook can see what the command produced
✓ does not run them in human mode, where oclif runs them itself
✓ never lets a broken hook take down the command
✓ publishes TRACEPARENT so postrun hooks and spawned processes share the trace
✓ adopts an inherited TRACEPARENT rather than replacing it
…
Tests: 13 passed
```

The fourth is the one to read carefully — running them *twice* in human mode would be the obvious way to get this wrong.

### 4. #199 — demo mode on the standalone BFF path

```bash
npx jest packages/bff-plugin --testPathPattern "mock-switch|mesh-config"
```

The gap was never in `bff:init`: `extractMeshConfig` yaml-dumps the manifest's `data:` block and dropped `mockSwitch` entirely, and the standalone path never renders `meshrc.yaml.ejs`. Check `applyMockSwitch` guards `responseCache` — without that, live and mock get cross-served from one cache, which is the failure mode you would not notice until it mattered.

### 5. Full gates

```bash
npm run check:adr && npm run build:adr-index
npm run build:docs && git diff --exit-code docs/api
```

Expect **139 suites / 2534 tests** and no diff from `build:docs` or `build:adr-index`.

### Known, out of scope here

`#318 cannot close` — three ADR-026 criteria need a retry-configuration decision (`errorHandling.retry.maxAttempts` has no field in `packages/dsl/src/schema.ts`), and inventing the field in an implementation PR would be the wrong order. Recorded on the issue and in ADR-026.

---

PR #342, opened 2026-07-27
feat(codegen): typed errors in generated code, and a fleet-wide regeneration drill

falese:
> **Stacked on #338.** Based on `claude/adr-dangling-issues-ir9eg8` so the diff is only this work; GitHub retargets it to `main` when #338 merges. Three files here are edits to files #338 introduces — the ADR-017 guard, `handlers/validation.ts`, and its generated doc — so it cannot stand alone against `main` without conflicting.

## What and why

#320 (in #338) typed every raw `throw new Error()` in the platform's own source. It left the templates, so **generated code was still entirely untyped** — 5 template sites fanning out to 48 files across 21 MFEs.

That was not an oversight anyone could have fixed by being careful. Generated MFEs import from `@seans-mfe-tool/runtime`, and that barrel exported `ValidationError` as a **type** from `context.ts` — the per-field *result record* `{field, message, expected?, actual?}` — while no error **class** appeared on it at all. An MFE author following ADR-017 had nothing to import.

`packages/runtime/src/errors/index.ts` already re-exported all six classes and described itself as a shim *"for one release cycle"*. It was simply absent from the barrel, and the blocker was the name collision.

## The change

**Making them reachable.** `context.ts`'s record is renamed `ValidationIssue` — it describes one failed field and is never thrown, and having "Error" in its name is what caused the collision. Two consumers, both updated. `handlers/validation.ts` drops the `ThrownValidationError` alias #320 had to introduce.

**What generated code now throws:**

| Emitted file | Was | Now |
|---|---|---|
| `platform/bff/bff.ts` | `Error` on non-2xx | `NetworkError(msg, response.status)` |
| `platform/bff/bff.ts` | `Error` on GraphQL errors | `BusinessError(msg, 'BFF_GRAPHQL_ERROR')` |
| `platform/base-mfe/mfe.ts` (React + Angular) | `Error` on unknown component | `ValidationError(msg, 'context.inputs.component', …)` |
| `src/index.tsx` | `Error` on missing root | `SystemError` |

**The BFF pair is the point.** `classifyError` (ADR-030) reads `type` to decide retryability, so every generated MFE's BFF failure — including a 503 — classified as `unknown` and non-retryable. The retry machinery could not help the one case it exists for:

```
raw Error   (before): {"type":"unknown","retryable":false}
NetworkError (after): {"type":"network","retryable":true}
GraphQL error (now):  {"type":"business","retryable":false}   ← correctly not retried
```

## The drill

Carrying this to the fleet doubled as a controlled test of a question the platform had never been asked: **how well does our own plumbing carry a breaking change?** Written up as a field report in the meridian style, at `docs/platform-design-review/breaking-change-regeneration-dx-report.md`.

**Where the machinery exists, it did well.** 😍 The drift gate caught all 29 generator-owned files on the first run, named each, printed the fix command, exit 1. `check:mfe-drift` fixed them in **3.9s** with nothing unrelated moving — the whitespace cascade CLAUDE.md warns about (PR #335) did not bite. And `remote:generate --force` over an already-regenerated MFE produced **byte-identical** output, demonstrating ADR-043's idempotence rather than asserting it.

**Where it stops, nothing measures the boundary.** 😕 Regeneration reached 29 of 48:

| Generated file | `overwrite` | Files | Reachable |
|---|---|---|---|
| `platform/base-mfe/mfe.ts` | `true` | 21 | ✅ |
| `platform/bff/bff.ts` | `true` | 8 | ✅ |
| `src/index.tsx` | `false` | 19 | ❌ |

With all 19 stale against their own template, `mfe:validate` returned `ok: true` and `check:mfe-consistency` exited 0. Neither is wrong — but between them the platform has no concept of a developer-owned file going stale. All 19 were byte-identical in the region that mattered; a human edited them because no tool could.

Filed from the punch list: **#339** (the reach gap), **#340** (`--dry-run` reports `create` for files it will skip), **#341** (three gate-hygiene items).

One correction worth recording: `--force` being a no-op is **not** a new finding — meridian punch-list item 21 already caught it and made its help text honest. That fix holds; the drill re-confirmed it. What is still live is the dry-run planner, a separate code path.

## Enforcement

The ADR-017 guard from #320 grows to cover `.ejs` templates and `examples/` — **528 files, 54 templates, 357 example files**. A decision the platform will not follow in the code it emits is not a platform decision. `src/codegen/templates/api` is excluded with its reason in a comment: that generator emits a standalone Express app with no path to the contracts, and ADR-063 has the axis parked behind #251.

New codegen assertions (`packages/codegen/src/__tests__/generated-typed-errors.test.ts`) pin the class **and its arguments** — a test checking only "not a raw Error" would pass on the wrong class and miss the retryability point entirely.

## Verification

- 140 suites, **2540 tests**, lint, typecheck, `build`
- `check:mfe-drift:check` and `check:mfe-consistency` clean; no schema drift
- `docs:check-links`: 1174 links, 0 broken
- `docs/api` regenerated (the barrel is a public surface)
- Built runtime verified to export all six classes: `NetworkError('…', 503)` → `.type = network`, `.retryable = true`

Refs #320, ADR-017, ADR-030, ADR-034, ADR-043


falese:
## Local-test runbook — `21d3631`

```bash
git fetch origin && git checkout claude/typed-errors-in-templates
npm ci
npm run build
```

### 1. The before/after that motivates the whole PR

Run this on the **base** first, so the contrast is yours rather than mine:

```bash
git checkout 8c2fc11                                  # #338's head, before this PR
grep -n "throw new" examples/abc-kids/flappy/src/platform/bff/bff.ts
```

```
39:    throw new Error(`BFF request failed: ${response.status} ${response.statusText}`);
45:    throw new Error(json.errors.map((e) => e.message).join('\n'));
```

The status code is *right there in the message string* and structurally unavailable. Now come back:

```bash
git checkout claude/typed-errors-in-templates
grep -n "throw new\|^import" examples/abc-kids/flappy/src/platform/bff/bff.ts | head -3
```

```
12:import { NetworkError, BusinessError } from '@seans-mfe-tool/runtime';
41:    throw new NetworkError(
50:    throw new BusinessError(
```

### 2. Why that matters — the retryability check

`classifyError` (ADR-030) reads the error's `type` to decide retryability. Save as `/tmp/classify.ts`:

```ts
import { NetworkError, BusinessError } from './packages/contracts/src/index';
import { classifyError } from './packages/contracts/src/error-classifier';

const cfg = { types: [], defaultBehavior: { retryable: false } } as never;
for (const e of [new Error('raw'), new NetworkError('BFF failed', 503), new BusinessError('gql', 'BFF_GRAPHQL_ERROR')]) {
  const c = classifyError(e, cfg);
  console.log((e.constructor.name + ':').padEnd(15), JSON.stringify({ type: c.type, retryable: c.retryable }));
}
```

```
$ npx ts-node --compiler-options '{"module":"commonjs"}' /tmp/classify.ts
Error:          {"type":"unknown","retryable":false}
NetworkError:   {"type":"network","retryable":true}
BusinessError:  {"type":"business","retryable":false}
```

Read all three rows, not just the middle one. Every generated MFE's BFF failure — **including a 503** — used to land on row 1, so ADR-030's retry machinery could not help the one case it exists for. Row 3 is the other half: a GraphQL error is a business fault and must stay non-retryable. A change that made everything retryable would have looked like a fix and been a worse bug.

### 3. Regeneration is idempotent — check it, don't take my word

```bash
npm run check:mfe-drift:check        # expect: 21 MFE(s) checked — no generator-owned drift
cd examples/abc-kids/flappy && node ../../../bin/run.js remote:generate --force && cd -
npm run check:mfe-drift:check        # still clean — the two entry points agree byte-for-byte
```

Two things to know before you read the output:

- **`--force` is a documented no-op.** Meridian punch-list item 21 already made its help text honest. It does not touch `src/index.tsx`, and `options.force` is correctly read nowhere in `unified-generator.ts`.
- **`remote:generate` leaves an untracked `.gitignore`** the drift gate does not manage. That is #341 item 3, not this PR. `git status` after, and delete it.

Angular MFEs print `WARNING: Missing template for public/demo.html` and `favicon.ico` on every drift run. Pre-existing, filed as #341 item 2, and genuinely annoying on a gate whose output you are meant to read — during the drill they sat between the STALE entries.

### 4. The enforcement grew teeth

```bash
npx jest src/__tests__/no-raw-error-throws.test.ts
npx jest packages/codegen/src/__tests__/generated-typed-errors.test.ts
```

The guard now covers **528 files, 54 templates, 357 example files** — `.ejs` included. A decision the platform will not follow in the code it emits is not a platform decision.

The codegen assertions pin the **class and its arguments**, deliberately: a test checking only "not a raw `Error`" passes on `new BusinessError(...)` in the BFF and misses the retryability point entirely. `expect(bff).toContain('response.status')` is the line doing the real work.

### 5. Full gates

```bash
npm run docs:check-links
npm run build:docs && git diff --exit-code docs/api
```

**140 suites / 2540 tests**, 1174 links / 0 broken, no `docs/api` diff.

### 6. What the drill found, which you can reproduce

The interesting result is a negative one. At the point 19 developer-owned `src/index.tsx` files were stale against their own template:

```
$ seans-mfe-tool mfe:validate examples/abc-kids/flappy --json
{"ok": true, "issues": []}

$ npm run check:mfe-consistency
21 MFE(s) checked — all internally consistent.      (exit 0)
```

Every gate green over 19 known-stale files, found by grep. Neither gate is wrong — that is the point, and it is why #339 exists. Full write-up in `docs/platform-design-review/breaking-change-regeneration-dx-report.md`; the fix is #343, stacked above this.

---

PR #343, opened 2026-07-27
feat(codegen): warn when developer-owned code uses something the platform changed (ADR-082)

falese:
> **Stacked on #342**, which is stacked on #338. Each retargets automatically as the one below it merges. Merge order: #338 → #342 → this.

Closes #339 — the headline finding from the regeneration drill in #342.

## The gap

Carrying ADR-017 to the fleet left **19 developer-owned `src/index.tsx` files stale** against the template that seeded them. Every gate stayed green:

```
$ seans-mfe-tool mfe:validate examples/abc-kids/flappy --json
{"ok": true, "issues": []}

$ npm run check:mfe-consistency
21 MFE(s) checked — all internally consistent.      (exit 0)
```

A human found them by grepping. All 19 were byte-identical in the region that mattered.

The ownership split is right — the generator must not rewrite files it seeded but does not own, *"even with `--force`"*. What was missing is the platform's ability to say the one thing that needed saying: **this file is yours, and something it uses has changed.**

## What this adds

**A declared registry** — `packages/codegen/src/platform-migrations.ts`. One entry per platform change that reaches seeded-but-unowned code: `id`, `since`, `failsAt`, a matcher, the `message`, the `fix`, the ADR.

Hand-written on purpose. A template diff cannot tell a contract change from a comment reflow, and cannot produce the `fix` at all — which is the half a developer actually needs. Two entries, both real, both from the commits immediately below this one: the raw-throw rule and the `ValidationError` → `ValidationIssue` rename.

**Usage, not staleness.** The rule asks whether code *uses* the thing that changed, not whether the file predates it. Staleness would need a per-file template revision, and would then flag anyone who legitimately rewrote their own file — forever. Usage reports exactly who is affected and goes quiet when they fix it, including by rewriting the file entirely.

**Warn, never rewrite.** This needed severity to exist at all: `ValidationIssue` had none and `ok = issues.length === 0` made every rule fatal. Severity defaults to `error`, so all six existing rules are untouched, and `ok` is now keyed to errors alone.

**Advisory until a declared version.** Each entry names the version at which its warning becomes an error, compared against the running CLI — there is no per-MFE version to read, and the ADR says so rather than inventing one. `--strict` escalates on demand.

## Reported where the developer already is

`mfe:validate`:

```
  ⚠ platform-migrations
      - Throwing a raw `Error` — the platform classifies failures by error type,
        so this is reported as `unknown` and never retried (ADR-017) src/index.tsx:79
        fix: Throw a typed error from '@seans-mfe-tool/runtime': ValidationError …

abc-kids-flappy is consistent, with 1 platform-migration warning(s).
```

`check:mfe-consistency` inherits it (it calls `mfeValidateCommand`) and now says so **in the summary**, not only inline — reporting "all consistent" over a fleet with known-stale code is the exact blindness this PR ends, and reproducing it in the gate's last line would be perverse:

```
21 MFE(s) checked — all internally consistent, with 1 platform-migration warning(s) in 1 MFE(s):
  ⚠ examples/abc-kids/flappy (1)
```

`remote:generate`, immediately after listing the files it declined to touch — the moment the gap is most visible:

```
Kept (developer-owned):
  src/index.tsx
  Yours to edit — regeneration never overwrites these

⚠ 1 of your file(s) use something the platform changed:
  src/index.tsx:79  Throwing a raw `Error` … (ADR-017)
      fix: …
```

Warnings also mirror into the envelope's `warnings[]` (ADR-018), so MCP tool calls and agents see them without parsing stdout.

## Design notes worth reviewing

- **Scoped to developer-owned files** by asking `generateAllFiles` — the same authority `check-mfe-drift` uses. A violation in a generator-owned file is a *drift* failure; reporting it in two places would teach people to ignore both. Costs ~0.5s per MFE.
- **Text matching, not an AST.** Consistent with `slots-implemented` and `parseFederationSharedEntries`. It will miss an alias. The cost of a false positive is a dismissed warning; the cost of the AST toolchain is paid on every run of every gate. Recorded in the ADR's Boundaries.
- **The ADR-017 guard grew a suppression marker** (`adr-lint-ignore: no-raw-error-throws`, honoured on the line or the one above, reusing the convention `adr-validation.ts` already has) — because the guard scans the registry that *describes* the rule. Prose about a rule is not a violation of it. The `typed-errors` entry is additionally worded to avoid the literal, so it needs no suppression of its own.
- **The registry is a convention with tooling behind it, not a guarantee.** A breaking change that forgets to declare an entry is as invisible as before. The ADR states that as an accepted cost rather than implying the mechanism enforces its own use.

## Verification

Reintroduced the exact defect the drill found, in a developer-owned file:

| | result |
|---|---|
| `mfe:validate` | warns; `ok` stays `true` |
| `check:mfe-consistency` | reports it, exit 0 |
| `mfe:validate --strict` | exit 65 |
| `remote:generate` | names it under "Kept (developer-owned)" |
| after `git checkout` | silent again, 0 warnings |

Gates: 141 suites, **2567 tests**, lint, typecheck, `build`, `check:adr`, `build:adr-index`, `check:mfe-drift:check`, `check:mfe-consistency`, `build:docs`, `docs:check-links`. `schemas/mfe-validate.json` regenerated for `severity`, `location`, `fix` and `--strict`.

Refs ADR-082, ADR-017, ADR-018, ADR-043, ADR-073

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tro4E4Pd4tzX5Nm2uhc6ay)_


falese:
## Added in `6d00999` — the convention half

The PR body already conceded the mechanism's limit: *"a convention with tooling behind it, not a guarantee"* — a hand-maintained registry only works if declaring an entry is part of making the change, and nothing made it so. This commit writes that convention down in the two places the change actually gets made.

### CLAUDE.md — a new section

Three trigger paths, with the third being the non-obvious one:

- `packages/codegen/templates/**`
- `packages/runtime/src/**` — the only package generated code imports from (13 of 13 template imports)
- `packages/contracts/src/**` — reaches MFEs through `packages/runtime/src/errors/index.ts`, a pure re-export shim, so a rename in contracts moves generated code **with no runtime file in the diff**

Then the two consequences and which one regeneration actually fixes, with the drill's numbers as the evidence: 48 files touched, 29 reached, 19 stale with every gate green. And the rule — the `PLATFORM_MIGRATIONS` entry ships in the same commit, not as a follow-up issue.

### The governance report asks on the PR

`scripts/adr-governance-report.ts` already computes `changedFiles()` and posts one marked comment, so this is a new section in it — no new job, no workflow change:

> ### ⚠️ This PR changes contracts that reach generated code
>
> `packages/codegen/templates/base-mfe-angular/mfe.ts.ejs` (+5 more)
>
> Regeneration re-stamps the files the generator owns (`npm run check:mfe-drift`). It **cannot reach developer-owned files** — `src/index.tsx`, `App.tsx`, anything an MFE author wrote — even with `--force`. Carrying ADR-017 out to the fleet reached 29 of 48 files and left 19 stale with every gate green.
>
> If this change affects code an MFE author wrote, declare an entry in `packages/codegen/src/platform-migrations.ts` (ADR-082) in this PR — or say here why none is needed.

**Suppressed when the diff already touches the registry** — a PR that did the right thing stays quiet. Keyed on the file rather than on added lines: the registry has exactly one purpose, so touching it *is* the declaration, and the check stays a pure function over the file list.

Advisory, matching the report's existing posture (ADR-075 §7 keeps live GitHub state in a comment rather than the correctness path). A blocking gate would fire on comment reflows and typo fixes, and an escape hatch used every third PR stops being read.

**Verified against the real ranges, not just fixtures:**

| Range | Result | Why |
|---|---|---|
| `origin/main..HEAD` (this PR) | silent ✅ | the diff declares two entries |
| `8c2fc11..21d3631` (#342) | fires, names 6 files ✅ | the commit that left the 19 files behind |

### Two corrections the drill proved

1. **`npm run typecheck` does not compile `packages/runtime`** — only `npm run build` does. The gate list implied otherwise, and this cost a `BusinessError` with a missing `code` argument twice. Takes the second option [#341](https://github.com/falese/seans-mfe-tool/issues/341) item 1 offers ("say plainly in CLAUDE.md"); the project-references option stays open there.
2. **`check:mfe-consistency` joins the pre-push gates.** CI-gated since #296, absent from the list, and now carrying this PR's warnings.

### One thing this turned up

`main()` was running at import, so importing the module ran a report over whatever `argv` happened to hold — the first test failed with `fatal: Not a valid object name scripts/__tests__/adr-governance-report.test.ts`, git being handed the test path as a ref. Guarded with `require.main === module`, which lets the script have its **first tests** after carrying a diff scan, blockquote stripping and table-cell escaping untested since ADR-075. `jest.config.js` learns to look under `scripts/`.

14 tests, and the balance is deliberate: 5 for the trigger, 6 for the silences. A reminder that appears on every PR is one nobody reads, at which point the registry goes unmaintained exactly as if the section did not exist.

**Gates:** 142 suites / **2581 tests**, lint, typecheck, build, `check:adr`, `build:adr-index` (no index diff), `check:mfe-drift:check`, `check:mfe-consistency`, `docs:check-links` (1176 links), `schemas/` clean.

ADR-082's Consequences section is updated to point at where the convention lives — the decision is unchanged, it just no longer stops at naming the gap.

---

falese:
## Local-test runbook — `6d00999`

Every block below is real output from this SHA.

```bash
git fetch origin && git checkout claude/warn-on-platform-migrations
npm ci
npm run build          # dist/ is what bin/run.js runs; rebuild or you test the old code
```

### 1. Reintroduce the exact defect the drill found

The honest test of a warning is whether it fires on the real thing, so put the real thing back:

```bash
cp examples/abc-kids/flappy/src/index.tsx /tmp/index.tsx.bak
sed -i "79s/.*/  throw new Error('Root element not found');/" examples/abc-kids/flappy/src/index.tsx
```

That is line-for-line what 19 files carried, in a **developer-owned** file — the class regeneration cannot reach.

### 2. `mfe:validate` — warns, and stays `ok`

```bash
node bin/run.js mfe:validate examples/abc-kids/flappy; echo "EXIT=$?"
```

```
  ✓ react-pinned
  ✓ manifest-package-sync
  ✓ shared-declared
  ✓ shared-version-sync
  ✓ runtime-declared
  ⚠ platform-migrations
      - Throwing a raw `Error` — the platform classifies failures by error type, so this is reported as `unknown` and never retried (ADR-017) src/index.tsx:79
        fix: Throw a typed error from '@seans-mfe-tool/runtime': ValidationError (bad input), BusinessError (precondition), NetworkError (transport, carries the status), SystemError (environment), SecurityError, TimeoutError

abc-kids-flappy is consistent, with 1 platform-migration warning(s).
EXIT=0
```

Check three things: the **file and line** (`src/index.tsx:79`), the **fix** — actionable without opening the ADR, which is the half no diff can generate — and **`EXIT=0`**. Warn, never rewrite. The file is untouched; confirm with `git diff --stat`.

```bash
node bin/run.js mfe:validate examples/abc-kids/flappy --strict >/dev/null 2>&1; echo "EXIT=$?"   # → 65
```

`--strict` is for a team that wants the contract before the platform requires it. Independent of the entry's `failsAt`, which escalates on a platform upgrade instead.

### 3. The fleet gate says so *in the summary*

```bash
npm run check:mfe-consistency; echo "EXIT=$?"
```

```
21 MFE(s) checked — all internally consistent, with 1 platform-migration warning(s) in 1 MFE(s):
  ⚠ examples/abc-kids/flappy (1)

These do not fail the gate. Run: seans-mfe-tool mfe:validate <dir>  (--strict to enforce)
EXIT=0
```

The summary line is the point, not the inline warning. Printing "all internally consistent" as the last thing a reader sees over a fleet with known-stale code is the exact blindness this PR ends; reproducing it in the gate's closing line would be perverse.

### 4. `remote:generate` names it where the gap is most visible

```bash
cd examples/abc-kids/flappy && node ../../../bin/run.js remote:generate; cd -
```

```
⚠ 1 of your file(s) use something the platform changed:
  src/index.tsx:79  Throwing a raw `Error` — the platform classifies failures by error type, so this is reported as `unknown` and never retried (ADR-017)
      fix: Throw a typed error from '@seans-mfe-tool/runtime': ValidationError (bad input), …
  Not changed for you. Run `seans-mfe-tool mfe:validate` for the full picture.

Summary:
  Generated: 16
  Preserved: 3
  Skipped: 10
```

Immediately after the files it declined to touch — the moment "yours to edit" and "the contract moved" are both true and visible at once.

⚠️ This leaves an untracked `.gitignore` (#341 item 3, pre-existing). `rm examples/abc-kids/flappy/.gitignore`.

### 5. It goes quiet when fixed

```bash
cp /tmp/index.tsx.bak examples/abc-kids/flappy/src/index.tsx
rm -f examples/abc-kids/flappy/.gitignore
node bin/run.js mfe:validate examples/abc-kids/flappy | grep platform-migrations   # → ✓ platform-migrations
git status --short                                                                 # → clean
```

Detection is by **usage**, not staleness — which is why it also goes quiet for a developer who fixed it by rewriting the file entirely. Staleness would have needed a per-file template revision and would flag that developer forever.

### 6. The PR-comment half

```bash
npx ts-node scripts/adr-governance-report.ts origin/main | grep -A3 "reach generated code"
```

**Nothing** — correct. This branch's diff touches `platform-migrations.ts`, so a PR that declared its entries stays quiet. Now the range that *should* fire, #342, the commit that left 19 files behind:

```bash
git diff --name-only 8c2fc11..21d3631 > /tmp/files.txt
npx ts-node -e "
const {generatedContractSection} = require('./scripts/adr-governance-report');
console.log(generatedContractSection(require('fs').readFileSync('/tmp/files.txt','utf8').split('\n').filter(Boolean)).join('\n'));"
```

```
### ⚠️ This PR changes contracts that reach generated code

`packages/codegen/templates/base-mfe-angular/mfe.ts.ejs` (+5 more)

Regeneration re-stamps the files the generator owns (`npm run check:mfe-drift`). It **cannot reach
developer-owned files** — `src/index.tsx`, `App.tsx`, anything an MFE author wrote — even with
`--force`. Carrying ADR-017 out to the fleet reached 29 of 48 files and left 19 stale with every
gate green.

If this change affects code an MFE author wrote, declare an entry in
`packages/codegen/src/platform-migrations.ts` (ADR-082) in this PR — or say here why none is needed.
```

That second invocation works only because `main()` is now guarded by `require.main === module` — before this commit, importing the module ran a report over whatever `argv` held.

### 7. Full gates

```bash
npm run check:adr && npm run build:adr-index && git diff --exit-code docs/spec.md docs/architecture-decisions/README.md
npm run docs:check-links
npx jest scripts/__tests__/adr-governance-report.test.ts    # 14 tests, the script's first
```

**142 suites / 2581 tests**, 1176 links / 0 broken, no index diff, `schemas/` clean.

One caveat on gate 2: `npm run typecheck` covers neither `packages/runtime` nor `scripts/` — `tsconfig.json` includes only `src/**/*`. `npm run build` and the jest run are what cover the code this PR touches.

---

PR #344, opened 2026-07-27
fix: the drill punch list, a React leak in the runtime barrel, and the Docker build path

falese:
> **Retitled and rewritten.** This opened as the #340/#341 punch list and grew four more commits while a human tried to actually compose a fleet with it. The description below covers all seven. Now targets `main` directly — #338, #342 and #343 have merged.

Closes #340, closes #341.

Everything here was reproduced before being fixed, and three of the fixes were confirmed by making a gate **fail** rather than watching it pass.

---

## 1. The regeneration drill's punch list (#340, #341)

**#340 — the dry run misdescribed the wet run.** The planner derived its operation from `file.overwrite` alone; the writer decides from existence *and* ownership:

| exists | `overwrite` | writer does | planner said |
|---|---|---|---|
| no | either | writes it | `create` ✅ |
| yes | `true` | re-stamps it | `overwrite` ✅ |
| yes | `false` | **skips it** | `create` / `(new)` ❌ |

Writing the test first turned up a second case the issue never mentioned: a generator-owned file that does *not* exist yet was announced as `overwrite`. Same expression, second lie. Against a real MFE, `{'overwrite': 16, 'skip': 10}` — all ten of those `skip`s previously read `create`.

**#341 item 1 — `typecheck` now covers `packages/runtime`.** It was compiled only by `npm run build`, so `typecheck` passed on runtime code that does not compile. Proved by reintroducing the exact defect: old gate `EXIT=0`, new gate `retry-wrapper.ts(265,11): error TS2554`.

**#341 item 2 — Angular template warnings, 8 → 0.** `demo.html`/`favicon.ico` are variant-owned now, following the slots probe beside them (ADR-036). `OPTIONAL_PUBLIC_ASSETS` is asserted to *exclude* `index.html` — silencing the diagnostic wholesale would trade a noisy gate for a blind one.

**#341 item 3 — `.gitignore` is generator-owned**, so `check:mfe-drift` covers it. All 7 committed copies were verified byte-identical to the template *before* anything was written; regeneration modified none of them. Ownership is per-entry — `package.json` stays developer-owned, pinned by a test.

## 2. A React import leaked into the runtime barrel 🐛

**The one that mattered.** Three Angular MFEs would not bundle:

```
./node_modules/@seans-mfe-tool/runtime/remote-mfe.js:133:60-87
  Error: Can't resolve 'react-dom/client'
```

`index.ts` re-exported `RemoteMFE`, so the barrel every generated MFE imports reached React transitively. ADR-056 *permits* `RemoteMFE` to import React — it is layer 5, it produces the native handle — but the barrel re-export leaked that permission to every consumer.

It hid because every consumer used `import type`, which erases at compile time and emits no require. It surfaced the day commit 1 of #342 added the first **value** import to an Angular-emitted template. Importing a typed error class started pulling React into MFEs that do not have it.

`RemoteMFE` now lives behind `@seans-mfe-tool/runtime/react`, mirroring `/angular`. That asymmetry — one abstract on the barrel, one behind a subpath — *was* the defect. Verified where it counts: `dist/runtime/index.js` no longer requires `remote-mfe` at all.

### The gate that could not fail

ADR-056 cites its boundary scan as *"the bright line made machine-checked"*. The pattern was:

```js
/(?:from|require\(\s*)['"](react|…)['"]/
```

No `\s*` before the quote, no `import(` alternative. It matched `require('react')` and nothing else — not `from 'react'`, not `await import('react')`. Neither form appears in this codebase. **It had never been able to fail**, and was cited as evidence throughout.

Repaired, and joined by a reachability walk over the barrel's module graph that prints the import *trail* rather than the leaf, because the fix is to cut an edge of the path.

Recorded as a section in **ADR-056**, not a new ADR — nothing new was decided, the boundary was violated. Follows ADR-075 §7's precedent for appending to a ratified decision. A `PLATFORM_MIGRATIONS` entry (`remote-mfe-subpath`) reports it for consumers outside this repo, matching value imports only.

## 3. The Docker build path

**`.dockerignore` is now generated.** Exactly one MFE of 21 lacked it — and that is the one that failed, on `COPY . .` trying to replace the staged runtime directory with the host's `node_modules` symlink. The file was emitted by nothing, hand-maintained, and had drifted into four shapes. Now generated with `.gitignore`, `overwrite: true`, drift-gated.

**Shared npm cache mount (#347).** A sequential fleet build died with `ENOSPC` because every builder stage carried its own copy of the ~850-package npm cache. The install step now writes to a shared BuildKit cache mount, never into the layer. Template plus all 18 existing Dockerfiles by hand — `Dockerfile.ejs` is `overwrite: false`, so regeneration reaches none of them.

**CLAUDE.md corrections**, both of which cost real debugging time:
- `turbo` is a devDependency, so the documented commands need `npx`.
- The Commands table pointed at `turbo run docker:build:examples`, which is a bare `docker compose build` — **13 services in parallel**, an OOM on any laptop, surfacing as an unrelated `npm install` exit 1 in whichever service loses. Both fleets already ship sequential builders that exist to prevent exactly this; neither was mentioned. Now they lead, with the turbo row kept and labelled.

## 4. What this found in the directives themselves

Asked whether the `.gitignore` ownership flip was a platform change, the honest answer was **yes — and the ADR-082 machinery shipped one commit earlier said nothing.** Verified before claiming it: `generatedContractSection(…)` → `SILENT`.

The trigger paths covered what generated code *contains* and what it *imports*, but not the generator logic deciding **who owns** what is emitted. `unified-generator.ts` is now a fourth trigger, scoped to that one file since it is the only non-test file that *sets* the flag.

An ownership change is the more dangerous edit: a template change alters content a reviewer sees; `overwrite: false → true` alters whether a developer's own edits survive.

## ⚠️ Costs, stated because nothing else can state them

- **A customised `.gitignore` outside this repo is overwritten** on next `remote:generate`. ADR-082 cannot warn — the instant the flag flips, the file leaves the registry's scope.
- **`RemoteMFE` imported as a value from the barrel is a breaking change.** The migration entry reports it with the fix; type-only imports were never broken and are not flagged.

## Verification

lint · typecheck (both projects) · **145 suites, 2618 tests** with coverage thresholds · `build` · `check:adr` (82 ADRs) · `build:adr-index` · `check:mfe-drift:check` · `check:mfe-consistency` · `slots:validate` (both fleets) · `build:schema:dsl:check` · meridian regen invariant · **`check:template-typecheck`** (fresh scaffold, real 1319-package install, real `tsc`, both React and Angular lanes) · `docs:check-links` · `schemas/` clean · `docs/api` regenerated.

Three gates confirmed by making them fail:

| Gate | Made to fail by | Result |
|---|---|---|
| `npm run typecheck` | dropping a `BusinessError` `code` arg | `TS2554` (old gate: exit 0) |
| `check:mfe-drift:check` | hand-editing a `.gitignore` | `STALE`, exit 1 |
| ADR-056 boundary scan | the barrel reaching `remote-mfe` | fails with the import trail |

And the thing CI still cannot do: **a human built and composed the fleet against this branch.** That is what found §2 — see #345.

## Follow-ups filed

**#345** (nothing in CI composes or bundles an MFE — now carrying §2 as evidence), **#346** (generated Dockerfiles ignore the committed lockfile), **#347** (meridian has no shared dependency base; only the cache-mount half is fixed here).

Refs ADR-056, ADR-036, ADR-043, ADR-082, ADR-075


falese:
## Local-test runbook — `c93dc3f`

Every block below is real output from this SHA.

```bash
git fetch origin && git checkout claude/drill-punch-list
npm ci
npm run build          # dist/ is what bin/run.js runs
```

⚠️ **`npm run build:packages` matters here more than usual.** `check:mfe-drift` resolves `@seans-mfe/codegen` to `packages/codegen/dist`, so without a rebuild you will be checking the *previous* build of the generator and item 2 will look unfixed. This bit me mid-PR; it is now documented in CLAUDE.md gate 8. `npm run build` includes it.

### 1. #340 — the dry run now matches the writer

In an MFE that has been generated before:

```bash
cd examples/abc-kids/flappy
node ../../../bin/run.js remote:generate --dry-run --json 2>/dev/null \
  | python3 -c "
import json,sys
from collections import Counter
d=json.load(sys.stdin)['data']['plannedChanges']
print('ops:', dict(Counter(p['op'] for p in d)))
for p in d:
    if p['target'] in ('src/index.tsx','package.json','.gitignore'): print(f\"  {p['op']:10} {p['target']}\")"
```

```
ops: {'overwrite': 16, 'skip': 10}
  skip       package.json
  overwrite  .gitignore
  skip       src/index.tsx
```

**All ten `skip`s previously read `create`.** Check the human output too — it is what a developer actually reads before deciding to run it:

```bash
node ../../../bin/run.js remote:generate --dry-run 2>/dev/null | grep -E "index.tsx|package.json|gitignore"
```

```
  package.json (skip — yours)
  .gitignore (overwrite)
  src/index.tsx (skip — yours)
```

`.gitignore` reading `overwrite` here is item 3 working, not a bug.

### 2. #341 item 1 — prove the typecheck gate by breaking it

Do not just watch it pass. Drop the second argument from the `BusinessError` at `packages/runtime/src/retry-wrapper.ts:265`, then:

```bash
npx tsc -p tsconfig.typecheck.json >/dev/null 2>&1; echo "old gate EXIT=$?"
npm run typecheck 2>&1 | grep "error TS"
```

```
old gate EXIT=0
packages/runtime/src/retry-wrapper.ts(265,11): error TS2554: Expected 2-3 arguments, but got 1.
```

The old gate is green on code that does not compile. That is the bug, reproduced. `git checkout` the file after.

Also worth confirming it needs no prior build — `rm -rf packages/*/dist && npx tsc -p packages/runtime/tsconfig.typecheck.json` should still pass, because the config inherits the root `paths` and resolves contracts from source.

### 3. #341 item 2 — the warnings are gone

```bash
npm run check:mfe-drift:check 2>&1 | grep -c "WARNING: Missing template"
```

`0`. Before: `8` (two per Angular MFE). Then confirm the diagnostic is not simply switched off:

```bash
npx jest packages/codegen/src/__tests__/optional-public-assets.test.ts
```

```
✓ emits no demo.html and no favicon.ico
✓ says nothing about the assets the Angular variant does not ship
✓ still emits both for React, which does ship them
✓ does not include index.html — an MFE with no HTML entry cannot be served
✓ is exactly the two assets base-mfe-angular legitimately lacks
```

The fourth is the one that matters: `index.html` must stay outside the optional set.

### 4. #341 item 3 — `.gitignore` is under the gate

The safety check first, which is what I ran before writing anything:

```bash
for f in $(git ls-files 'examples/*/*/.gitignore'); do
  diff -q "$f" packages/codegen/templates/base-mfe/.gitignore.ejs >/dev/null \
    && echo "IDENTICAL $f" || echo "DIVERGED  $f"
done
```

All seven MFE copies `IDENTICAL` — nothing was clobbered. (`api-examples/*` and `meridian-station/apis/*` show `DIVERGED`; those come from the API generator, have no `mfe-manifest.yaml`, and the drift gate never reaches them.)

Now make the gate fail:

```bash
echo "# custom line" >> examples/meridian-station/meridian-concourse/.gitignore
npm run check:mfe-drift:check 2>&1 | grep STALE; echo "EXIT=${PIPESTATUS[0]}"
git checkout examples/meridian-station/meridian-concourse/.gitignore
```

```
  STALE: examples/meridian-station/meridian-concourse/.gitignore
EXIT=1
```

That exit 1 **is** the accepted cost, seen directly: the platform will now overwrite this file. If you would rather it stayed developer-owned, this is the command that shows you what changes.

### 5. The directive gap this PR closes

The reminder shipped in #343 was silent on this very PR. Confirm the fourth trigger works:

```bash
git diff --name-only 6d00999..HEAD > /tmp/f.txt
npx ts-node -e "
const {generatedContractSection} = require('./scripts/adr-governance-report');
const f = require('fs').readFileSync('/tmp/f.txt','utf8').split('\n').filter(Boolean);
const o = generatedContractSection(f);
console.log(o.length ? o.slice(0,3).join('\n') : 'SILENT');"
```

```
### ⚠️ This PR changes contracts that reach generated code

`packages/codegen/src/unified-generator.ts`
```

Check out `b183c59` and re-run the same command to see `SILENT` — the state this PR fixes.

### 6. Full gates

```bash
npm run build:schemas && git diff --exit-code schemas/
npm run check:adr && npm run check:mfe-drift:check && npm run check:mfe-consistency
npm run docs:check-links
```

**145 suites / 2600 tests**, 1176 links / 0 broken, `schemas/` clean.

---

PR #352, opened 2026-08-15
@falese/smt-* namespace and registry-based package delivery (2/3)

falese:
**Stack 2 of 3 — targets #351, not `main`.** Merge #351 first.

| PR | Branch | Base |
|---|---|---|
| 1 | `claude/split-1-gates-and-demos` | `main` (#351) |
| **2 (this)** | `claude/split-2-packaging-namespace` | PR 1 |
| 3 | `claude/split-3-conformance` | PR 2 |

## This is the high-risk one

10 commits, **475 files — of which 235 are regenerated `examples/**` output** that has to travel with the rename. Governed by ADR-083 (namespace), ADR-084 (two delivery lanes), ADR-085 (compile contract).

⚠️ **Contains a breaking rename.** Every platform package moved to `@falese/smt-*`.

**The review question is not "is the rename correct in spirit"** — CI answers that. It is **"is there a specifier the sweep missed."** That is exactly what 475 files hides, and it is the reason this PR is separate.

## Why the namespace changed

GitHub Packages pins the npm scope to the repo owner, so `@seans-mfe/*` could not be published there at all, and third parties could not publish into it. ADR-083 supersedes ADR-021.

ADR-084 then gives two delivery lanes: GitHub Packages hosted, plus a static mirror baked into the CLI image — which preserves ADR-033's no-registry promise.

## Where the real logic changed

Everything else is rename + regeneration.

| File | What |
|---|---|
| `packages/codegen/src/lockfile.ts` | **new** — lockfile lane-independence |
| `scripts/build-registry-mirror.ts` | **new** — static registry mirror for the CLI image |
| `scripts/serve-registry-mirror.js` | **new** — serves it locally |
| `packages/codegen/src/validate.ts` | `runtime-declared` now rejects `file:` specifiers; adds `compile-contract-inherited` |
| `packages/codegen/src/platform-migrations.ts` | ADR-082 migration entries |
| `packages/codegen/src/unified-generator.ts` | generated slots bind the published `DeclaredSlot` |

## Two findings already fixed inside this range

Both are worth understanding before reviewing the rest, because both were invisible to the test suite:

**`npm ci` ignores `.npmrc` scoped registries.** It follows lockfile `resolved` URLs instead. An earlier "pass" turned out to be the npm cache masking it — caught only after clearing the cache. Fixed by stripping `resolved` while keeping `integrity`.

**Moving runtime output to `dist/` broke every generated MFE.** Under `moduleResolution: "node"` (Node10) the `exports` map is ignored entirely. Found only by re-running the demo end to end, because **jest uses `moduleNameMapper` and typecheck uses tsconfig paths — both bypass the published layout**. Fixed with root-level forwarders plus a test asserting on the shipped `files` list.

## What CI already covers

`check:mfe-drift:check` compares generated bytes across all 21 example MFEs; `check:mfe-consistency` checks manifest ⇄ package.json ⇄ federation agreement. A missed specifier in generated output fails those. A missed specifier in *hand-written* source is what review is for.

---
_Generated by [Claude Code](https://claude.ai/code/session_016ZcVaKqzVyEiQ8bU3y4vMn)_


falese:
## 🔴 `quality` is red — one fix I own, one decision I don't

`check:template-typecheck` fails on both lanes. Reproduced locally. Two separate causes.

### 1. The harness is stale — mine to fix

`scripts/check-template-typecheck.ts:86` generates the scaffold (which correctly emits `^0.1.0` per ADR-084), then **rewrites it back to the staging model**:

```ts
// `@falese/smt-runtime` isn't published to npm yet (ADR-064) — real
// Dockerfiles stage it via `npm pkg set ... file:.../dist/runtime`
pkgJson.dependencies['@falese/smt-runtime'] = `file:${DIST_RUNTIME}`;
```

That comment is no longer true — ADR-084 replaced staging with two registry lanes — and this PR adds the `runtime-declared` rule that **rejects `file:` specifiers**. So the harness now sets up exactly the state its own validation forbids:

```
✗ runtime-declared
    "@falese/smt-runtime" is declared as "file:/home/runner/.../dist/runtime" —
    a path specifier pins this MFE to one checkout layout and one delivery lane,
    which is the staging model ADR-084 replaced
```

A gate that wasn't migrated when the decision it validates changed. The fix is to install the probe from the offline mirror this PR builds (`scripts/serve-registry-mirror.js` + `dist/registry`) rather than rewriting to `file:`.

### 2. Angular regression — needs a call

```
✗ manifest-package-sync
    manifest implies dependency "@falese/smt-framework-angular" (^0.1.0)
    but it is absent from package.json
```

Traced to `029a511` in this PR. It added:

```ts
const FRAMEWORK_RUNTIME_PACKAGES = {
  react:   '@falese/smt-framework-react',
  angular: '@falese/smt-framework-angular',
};
```

`resolveClientDependencies` adds the framework package for whichever framework, and `validate.ts:227` uses that same function to compute implied dependencies. But:

- Angular's `package.json.ejs` is a separate template and never receives those deps
- `base-mfe-angular/slots.ts.ejs` **does not import `@falese/smt-framework-angular` at all**

So the "generated slots bind the published `DeclaredSlot`" change landed **React-only**, while the dependency table claims both. Net effect: **every fresh Angular scaffold fails `mfe:validate`.** Verified by generating both lanes — React declares `@falese/smt-framework-react`, Angular declares no `@falese/*` dependency at all.

The existing Angular example (`meridian-life-support`) passes `check:mfe-consistency` because its `package.json` was patched by hand earlier in the branch. Only the fresh-scaffold gate sees this — which is the gate's whole purpose, and I never ran it locally.

**Two defensible fixes, and they are different decisions:**

| | Fix | Implication |
|---|---|---|
| **a** | Make Angular a real peer — template binds the published directive, package.json declares the dep | Larger; matches ADR-084's stated intent |
| **b** | Restrict `FRAMEWORK_RUNTIME_PACKAGES` to `react` | Smaller and honest; defers the Angular half and says so |

I'm not picking between those — it's an architectural asymmetry, and CLAUDE.md says to ask rather than invent. **@falese** which way?

### Status of the other two PRs

- **#351** — `api-docs` was red for an unrelated reason and is **fixed** (`a5a1093`). TypeDoc omits the source link for a file that isn't tracked yet, so `control-plane-state.ts` was documented in the plain form before it was committed. Regenerating after the fact produces the linked form the gate expects.
- **#353** — same TypeDoc cause for `child-stdio.ts`, plus a `build:docs` I owed after the triage commit touched `adr-validation.ts`. **Fixed** (`fe4b9eb`). Its `quality` failure is this same Angular issue, inherited from here.

Tests are green everywhere: 2672 / 2713 / 2864 across the three PRs, both Node lines.

---

falese:
## Correction: there is a **third** cause, and it's the important one

My previous comment attributed the React lane's failure to the harness. That was wrong — the React lane never reaches validation. It dies at `npm install`:

```
npm error 404 Not Found - GET https://registry.npmjs.org/@falese%2fsmt-framework-react
npm error 404  '@falese/smt-framework-react@^0.1.0' is not in this registry.
```

### A freshly generated MFE cannot `npm install`

ADR-084's contract is "the manifest declares plain semver, and `.npmrc` chooses the lane." This PR shipped the semver half. **The `.npmrc` half only exists inside the Docker image.**

- `packages/codegen/templates/base-mfe/` emits **no `.npmrc`**
- no `examples/**` MFE has one
- `Dockerfile.cli:26` sets `@falese:registry` at the mirror — which is why the container path works
- and `scripts/build-registry-mirror.ts:123` already says *"Matches the .npmrc a generated MFE carries in the offline lane"* — describing a file generated MFEs **do not carry**

So `seans-mfe-tool remote:init` on a developer's machine produces a project that declares a dependency it cannot resolve, and `npm install` 404s against npmjs.org. The examples don't surface it because they resolve through the workspace; only the fresh-scaffold gate sees it, which is the gate's whole purpose.

This is squarely this PR's business — `029a511` is what made the scaffold declare `@falese/smt-framework-react` in the first place.

### The three causes, ranked

| # | Cause | Kind |
|---|---|---|
| **1** | No `.npmrc` emitted → fresh scaffold can't install | **ADR-084 half-implemented** |
| **2** | `manifest-package-sync` implies `@falese/smt-framework-angular`; Angular template neither declares nor imports it | regression from `029a511` |
| 3 | Harness rewrites runtime to `file:${DIST_RUNTIME}`, which its own `runtime-declared` rule now rejects | stale gate, mine |

### What I need from you

**(1) is a decision, not just a fix.** ADR-084 clearly intends a generated `.npmrc` — `build-registry-mirror.ts` documents one. But *which lane does a generated MFE default to?*

- **Hosted** — `@falese:registry=https://npm.pkg.github.com`. Matches the published story, but GitHub Packages requires a token even for public reads, so `npm install` needs auth on a fresh clone.
- **Offline mirror** — points at the CLI image's mirror. Works with no auth, but only where the mirror is reachable.
- **Both, with the hosted one default and the Dockerfile overriding** — probably the intent, worth confirming.

That choice sets the install story for every MFE this platform generates, so I'm not picking it.

**(2)** still needs the (a)/(b) call from my previous comment.

**(3)** I'll fix alongside whatever you choose.

Sorry for the mis-attribution — I read the harness rewrite and stopped before checking why React failed differently from Angular. The two lanes fail for two different reasons and I collapsed them.

---

falese:
## `quality`: install now works, typecheck doesn't — and I can't finish this responsibly right now

The `.npmrc` + mirror work did its job: **both lanes install**. `[react] OK` end to end. The Angular lane gets past install and fails later, on two things my local run did not surface.

### 1. The failure — `@falese/smt-runtime` subpaths don't resolve from an installed Angular MFE

```
mfe.ts(3,10):  Module '"@falese/smt-runtime/angular"' has no exported member 'AngularRemoteMFE'
mfe.ts(4,33):  Cannot find module '@falese/smt-runtime' or its corresponding type declarations
bootstrap.ts(40,4): Property 'load' does not exist on type 'templatetypecheckprobeMFE'
TS4112: ... does not extend another class
```

That last pair is the tell — the class silently has no base, because the import above it failed. This is the same **shape** as the Node10 `exports`-map problem fixed earlier in this branch with root-level forwarders (`da22586`), now appearing on the Angular subpath from a genuinely installed package.

**It passed locally and failed here**, which means my local run was not testing what CI tests — most likely a warm npm cache serving a tarball that predates one of today's package changes. I don't want to guess at the cause under time pressure; a wrong packaging fix is how the last two rounds of this went.

### 2. A design consequence I should have predicted

```
⚠ lockfile-lane-independent
    "@falese/smt-runtime" is locked to http://127.0.0.1:4873/@falese/smt-runtime/-/...tgz
```

Installing the probe through the mirror produces a lockfile pinned to the mirror URL — **exactly what ADR-084 §5 forbids**. It's a warning so it isn't the failure, but it means the gate now manufactures the defect its sibling rule exists to catch. The harness should strip `resolved` from `@falese/*` entries after install, the same normalisation ADR-084 §5 requires of generated MFEs. Otherwise the gate models a lane no consumer should ever ship.

### Where this leaves the PR

Everything else on #352 is green: `build`, `api-docs`, `adr-governance`, `e2e`, `perf`, `doc-links`, and 2713 tests on both Node lines. The `.npmrc`, the precedence rule, the Angular binding, the out-of-process mirror and the Angular peer range are all correct and staying.

What's left is one real bug (subpath resolution from an installed package, Angular lane) plus one harness correction (strip `resolved`). Both are tractable; neither should be attempted on a guess.

**@falese** — I'm at the end of what I can verify carefully in this session. Recommend picking this up fresh with the two items above as the brief. #351 remains fully green and mergeable independently of all of it.

---

PR #359, opened 2026-08-16
docs: rebuild the system map as an executive narrative, in literal language, and publish it to GitHub Pages

falese:
## What this does

Turns `docs/system-map.html` from a good architecture visualisation into an explanation of the platform's operating model, aimed at executives and product leaders first and engineers second — and publishes it.

Three changes, in order of importance:

1. **The business problem became the protagonist.** The page led with the architecture. It now leads with *why the platform exists*, and the architecture arrives as the answer.
2. **The metaphors are gone.** Brief, factory, house standard, hinge, control tower, rule book, placer, machinery, "the screen assembles itself" — nine metaphors for eight concepts, all replaced with literal terms. The reader no longer has to learn a vocabulary before understanding the system.
3. **It is published.** `https://falese.github.io/seans-mfe-tool/system-map.html`, via a workflow that is now the repository's single Pages publisher.

---

## The narrative

| Section | Answers |
|---|---|
| Why it exists | A without/with comparison: three teams, three capabilities, differing only in what sits between them |
| Four promises | Team autonomy · one platform standard · faster change · one customer experience |
| How it works | Capability definition → build system → platform standard → composition rules → runtime → customer experience |
| Follow one capability | Nine stages, two real capabilities, switchable |
| What the platform provides | Eleven responsibilities, grouped by function not folder |
| Detailed architecture | The existing map, relabelled as the deeper layer |
| Technical traceability | Plain-English term ⇄ repository term |
| What this enables / Trade-offs | 7 benefits, 5 costs, each `fact` or `inference` |
| Repository at a glance | The counts, moved out of the masthead |

The central message is stated outright rather than implied: **many teams, one platform standard, one product** — teams do not need to build everything the same way, they need to agree on the standard that lets their capabilities work together.

## Canonical vocabulary

Applied to every heading, diagram label, card, walkthrough stage, legend entry and `aria-label`:

| Was | Now |
|---|---|
| Brief | Capability definition |
| Factory | Build system |
| House standard | Platform standard |
| Hinge | Platform contract |
| Control tower | Composition rules |
| Rule book | Rules service |
| Placer | Layout manager |
| Router | Message channel |
| Empty shell | Product host |
| Machinery | Platform |
| Feature (noun) | Capability |

The internal CSS class `.hinge` was renamed `.standard-box` so the metaphor is gone from the source too. A grep for every retired term across the published pages returns zero.

**Technical names are preserved, not lost.** The detailed architecture layer uses the repository's own terms — control plane, registry, daemon, layout manager, `BaseMFE`, `providesSlots` — and the traceability section carries an explicit mapping, so an executive reads the function and an engineer reads the implementation of the same thing.

## Corrections found while working

Each of these was wrong on the previous page and is fixed here:

- **"The feature brief is the only source of truth"** → *the source of truth for what the capability is meant to do*, with an explicit note that it is not the source of truth for runtime state, live data, or what is deployed.
- **Product decision count: 11 → 8.** `PDR-001`–`008`.
- **The capabilities section claimed ten entries and rendered eleven.**
- **The authorisation claim was too generous.** `packages/runtime/src/handlers/auth.ts` does verify signatures (HS256), but the generated `doAuthorizeAccess` returns `true` and `mesh-context.js` decodes the caller's token without verifying it. All three are now stated separately, with the "designed seam, undelivered policy" reading labelled as inference.
- **Deciding what appears and composing it were one box.** Composition rules and runtime are separate components with separate failure modes; the six-step model now separates them.

## Deployment

`docs/system-map.html` becomes a complete standalone document (it was an HTML fragment) and gains a landing page at `docs/index.html`.

`.github/workflows/pages.yml` is new and is now the **single** Pages publisher, serving:

```
/                                    docs/index.html
/system-map.html                     docs/system-map.html
/slides/platform-architecture.html   built from docs/slides/
/slides/platform-architecture.pdf
```

`slides.yml` was previously publishing `docs/slides/dist/` as the entire site root. It now builds the deck as a PR artifact but **does not deploy** — two workflows uploading a Pages artifact would each overwrite the other's whole site. This moves the deck from `/platform-architecture.html` to `/slides/platform-architecture.html`; nothing in the repository linked to the old URL.

No build step was added for the map: it is self-contained, so it is copied verbatim — no bundler, no base-path rewriting, nothing to keep in sync.

`scripts/check-site.js` is new and runs in the workflow. It fails the build on root-relative links, unresolvable in-site links, or externally-loaded assets — the three ways a project page served from `/seans-mfe-tool/` breaks. Negative-tested against all three.

**GitHub Pages is already enabled with source = GitHub Actions** (verified: a `deploy-pages@v4` job succeeded on `main`, and the currently-published deck returns 200). No repository setting is required. The URL goes live when this merges.

## Verification

Served from a local `/seans-mfe-tool/` sub-path in real Chromium:

- [x] **No console errors, no failed requests, no 4xx** — inline data-URI favicon added so even the default favicon request does not 404
- [x] **Repository paths hidden by default**, revealed by the nav toggle in cards, SVG and walkthrough stages; both toggles sync in each direction
- [x] **Legend layer filter** dims non-focused layers (0.14 vs 1) and resets
- [x] **Capability switcher** swaps all nine stages, by mouse and by keyboard; works with JavaScript disabled (default selection is CSS-driven)
- [x] **No horizontal page overflow** at 1440 / 1024 / 820 / 768 / 414 px
- [x] **Dark mode** paints an explicit background and foreground
- [x] **No SVG text overflow** in any of the four diagrams — screenshotted and inspected after the longer literal labels
- [x] `node scripts/check-site.js _site` — 4 pages, 4 in-site links, 0 problems
- [x] `npm run docs:check-links` — 350 files, 1205 links, 0 broken
- [x] Both workflows parse as valid YAML; exactly one Pages publisher remains

**Not run, and why:** `lint` (scope is `src jest.setup.js`), `typecheck` (CLI tree + `packages/runtime`), `test`, `build`, `build:schemas`, `check:adr`, `check:mfe-drift`, `check:mfe-consistency`, `compose:validate`, `build:docs` — this change touches no `src/**`, no `packages/**`, no ADR, no template, no manifest and no composition document.

**ADR-082 / `PLATFORM_MIGRATIONS`:** no entry needed. This change touches no contract that reaches generated code — nothing under `packages/codegen/templates/**`, `packages/runtime/src/**` or `packages/contracts/src/**`.

**Shared files:** `.github/workflows/slides.yml` is modified — deliberately, since it was the conflicting Pages publisher and the conflict cannot be resolved without touching it. `package.json`, `pnpm-workspace.yaml`, `turbo.json` and `schemas/` are untouched.

## Limitations

- The Pages **deploy path is unproven in CI** — the job is `main`-gated and workflow dispatch returned 403, so I could not dry-run it. Its Marp steps are copied verbatim from the workflow that has been deploying successfully; the assembly and verification steps were run locally.
- The map's deployment is **coupled to the slide build**: if Marp fails, the job fails and nothing deploys. Marp has never failed in this repository's recorded history (the only two red `slides.yml` runs died at `Set up job` in ~2 seconds), but the coupling is real and easy to remove if preferred.
- The groupings in "What the platform provides" and the stage boundaries in the walkthrough are **editorial judgements**, stated as such in the ambiguities section.
- One claim on the page is genuinely soft — *"a new team can be productive without reading the platform's source"* — and is labelled `inference` with the reason inline: the repository shows the mechanism and 21 worked examples but contains no evidence about onboarding time for outside teams.

## Reviewing this

The diff on `docs/system-map.html` is large and not useful to read line by line. Suggested review path:

1. Open the file locally — `open docs/system-map.html` — it is self-contained.
2. Read the first three sections as if you had never seen the architecture. That is the test the page is built to pass.
3. Turn on **Show technical implementation** and check that the repository paths still point where you expect.

---

https://claude.ai/code/session_017WapT1REGEXJqK1FiVVj61

---
_Generated by [Claude Code](https://claude.ai/code/session_017WapT1REGEXJqK1FiVVj61)_

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -9,8 +9,14 @@
 # Published layout (site root = https://falese.github.io/seans-mfe-tool/):
 #
 #   /                              docs/index.html          — landing page
+#   /404.html                      docs/index.html          — fallback
 #   /system-map.html               docs/system-map.html     — the system map
 #   /slides/platform-architecture.html|.pdf                 — the deck
+#
+# The system map is the point of this site and does not depend on the deck: the
+# Marp steps are non-fatal, and scripts/assemble-site.js removes the deck card
+# from the landing page if the deck was not produced. A failed deck build is a
+# warning and a slightly smaller site, never a missing system map.
 #
 # Every published page is self-contained: no external CSS, JS, fonts or images,
 # and every in-site link is relative. Nothing here assumes the site is served
@@ -53,27 +59,34 @@ jobs:
         with:
           node-version: '20'
 
-      - name: Assemble the site
-        run: |
-          set -euo pipefail
-          mkdir -p _site/slides
-          cp docs/index.html      _site/index.html
-          cp docs/system-map.html _site/system-map.html
-          # 404 falls back to the landing page rather than GitHub's default.
-          cp docs/index.html      _site/404.html
-
+      # The deck needs a third-party CLI and a headless browser — the most
+      # failure-prone steps here. Both are non-fatal: if the deck cannot be
+      # built, the assemble step below drops its card from the landing page and
+      # the system map publishes anyway. The map never waits on the deck.
       - name: Install Marp CLI
+        id: marp-install
+        continue-on-error: true
         run: npm install -g @marp-team/marp-cli
 
       - name: Build slides (HTML + PDF)
+        id: slides
+        if: steps.marp-install.outcome == 'success'
+        continue-on-error: true
         run: |
           set -euo pipefail
+          mkdir -p _site/slides
           marp docs/slides/platform-architecture.md \
             --output _site/slides/platform-architecture.html \
             --html
           marp docs/slides/platform-architecture.md \
             --output _site/slides/platform-architecture.pdf \
             --allow-local-files
+
+      # Copies the map and landing page unconditionally, and keeps the deck
+      # card only if the deck actually exists — so the published site can never
+      # contain a link to a page that was not built.
+      - name: Assemble the site
+        run: node scripts/assemble-site.js _site
 
       # Fails the build on a broken in-site link or a reference to an external
       # host, either of which would 404 (or be blocked) once published.
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ constant; the delivery mechanism is a template variant.
 >
 > **New here?** Start with the [**system map**](https://falese.github.io/seans-mfe-tool/system-map.html)
 > — an interactive explanation of why the platform exists, how it lets many teams
-> build one product, and what happens to a single feature end to end.
+> build one product, and what happens to a single capability end to end.
 
 ---
 
@@ -313,8 +313,9 @@ one gets the depth it needs and no more:
 | **Engineering** | Where is each concept implemented? Toggle **Show technical implementation** to reveal repository paths throughout the page. |
 
 It opens with the business problem — many teams changing one product — rather than with
-the architecture, then works down through the factory/control-tower mental model, an
-interactive **Follow a feature** walkthrough of two real capabilities in `examples/`, the
+the architecture, then works down through the six-step model (capability definition →
+build system → platform standard → composition rules → runtime → customer experience), an
+interactive **Follow one capability** walkthrough of two real capabilities in `examples/`, the
 platform's capability map, the detailed architecture diagram, and finally what the
 architecture *enables* and what it *trades off*. Claims are labelled `fact` or
 `inference` so the evidence behind each one is checkable.
--- a/README.md
+++ b/README.md
@@ -345,16 +345,25 @@ assembles a `_site/` directory, verifies it, and deploys:
 
 ```
 /                                    docs/index.html         — landing page
+/404.html                            docs/index.html         — fallback for a bad URL
 /system-map.html                     docs/system-map.html    — the system map
 /slides/platform-architecture.html   built from docs/slides/ — the architecture deck
 /slides/platform-architecture.pdf
 ```
 
-Three things this deliberately does *not* do:
+Four things this deliberately does *not* do:
 
 - **No build step for the map.** `docs/system-map.html` is copied verbatim. It is
   self-contained by design, so there is no bundler, no base-path rewriting, and nothing
-  to keep in sync. (The slide deck still needs Marp, which the workflow installs.)
+  to keep in sync.
+- **The map never waits on the deck.** The deck needs Marp and a headless browser — the
+  most failure-prone steps in the pipeline — so both are non-fatal. If the deck cannot be
+  built, [`scripts/assemble-site.js`](./scripts/assemble-site.js) removes its card from
+  the landing page and the site publishes without it, with a workflow warning. A failed
+  deck build costs you the deck, never the system map, and never leaves a link to a page
+  that was not published. (The card is delimited in `docs/index.html` by
+  `<!-- deck-card:start -->` / `<!-- deck-card:end -->`; the script errors out if those
+  markers go missing rather than publishing a dangling link.)
 - **No root-relative paths.** The site is served from `/seans-mfe-tool/`, not `/`, so
   every in-site link is relative. [`scripts/check-site.js`](./scripts/check-site.js) runs
   in the workflow and fails the build on any root-relative link, unresolvable in-site
@@ -362,6 +371,17 @@ Three things this deliberately does *not* do:
 - **No second Pages publisher.** [`slides.yml`](./.github/workflows/slides.yml) builds
   the deck and keeps it as a workflow artifact for PR review, but does not deploy — two
   workflows uploading a Pages artifact would each overwrite the other's whole site.
+
+To reproduce either outcome locally:
+
+```bash
+# deck present
+mkdir -p _site/slides && echo '<html><title>deck</title></html>' > _site/slides/platform-architecture.html
+node scripts/assemble-site.js _site && node scripts/check-site.js _site
+
+# deck build failed
+rm -rf _site && node scripts/assemble-site.js _site && node scripts/check-site.js _site
+```
 
 Pull requests touching the site build and verify it, and upload a `docs-site-preview`
 artifact instead of deploying.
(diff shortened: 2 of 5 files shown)


falese:
## Local-test runbook

Accurate to `d511fd6`. No rebuild of `dist/` or any Docker image is needed — this change touches no `src/**` or `packages/**`.

### 1. Fastest check — open the file

```bash
open docs/system-map.html        # xdg-open on Linux
```

The page is fully self-contained (no external CSS, JS, fonts or images), so `file://` renders it exactly as it will be served.

### 2. Test it the way GitHub Pages will serve it

The site is served from `/seans-mfe-tool/`, not `/`. To reproduce that sub-path:

```bash
mkdir -p /tmp/site/seans-mfe-tool/slides
cp docs/index.html      /tmp/site/seans-mfe-tool/index.html
cp docs/system-map.html /tmp/site/seans-mfe-tool/system-map.html
npx http-server /tmp/site -p 8080 -c-1
```

Then open <http://localhost:8080/seans-mfe-tool/system-map.html>.

Landing-page link to the deck will 404 locally — the deck is built by Marp in CI, not committed. Everything else should resolve.

### 3. Verify the site checker

```bash
node scripts/check-site.js /tmp/site/seans-mfe-tool
# check-site: OK — 2 page(s), N in-site link(s), 0 asset reference(s), 0 problems
```

To confirm it actually catches problems rather than passing vacuously:

```bash
mkdir -p /tmp/bad && printf '<a href="/x.html">a</a><link rel="stylesheet" href="https://cdn.example.com/a.css"><a href="nope.html">b</a>' > /tmp/bad/i.html
node scripts/check-site.js /tmp/bad; echo "exit=$?"
# 3 problems, exit=1 — root-relative link, external asset, unresolvable link
```

### 4. Drive the interactions

| Do this | Expect |
|---|---|
| Load the page | No repository paths anywhere; no console errors |
| Toggle **Show technical implementation** (top bar) | Paths appear in cards, inside the SVG diagram, and at every walkthrough stage. The in-section **Show repository paths** switch flips with it |
| Toggle the in-section switch instead | The top-bar switch follows it back — they are one state |
| Click a legend chip (e.g. *Runtime composition*) | Other layers fade to 14%; chip gets `aria-pressed="true"`. **Show all** clears it |
| Click **A game called Letter Pop** | All nine walkthrough stages swap their concrete example; the status line under the picker updates |
| Tab to the picker and press Enter | Same result — it is keyboard-operable |
| Scroll | Progress bar fills; the current section highlights in the nav |
| Resize to tablet and phone widths | No horizontal page scroll. Wide diagrams scroll **inside** their own panel |
| Switch OS to dark mode | Explicit dark background and foreground, not a borrowed one |

### 5. The actual acceptance test

Read the first three sections as if you had never seen this architecture, then answer:

1. Why does this platform exist?
2. How does it let independent teams build one product?
3. What happens to one capability, from definition to customer?

If any of those needs a second read, the language still needs work — tell me which sentence.

### 6. If something looks wrong

- **Text overflowing an SVG box** → the literal terms are longer than the metaphors they replaced; the `tspan` line breaks are hand-set in `docs/system-map.html`. Tell me which box.
- **A repository path that no longer exists** → the traceability table (`§ Where each idea is implemented`) and the `.paths` lines are the two places paths live.
- **Page renders unstyled** → something introduced an external asset reference; `scripts/check-site.js` should have caught it, so that would be a checker bug worth reporting.

---

falese:
## Marp decoupled — the map no longer waits on the deck

Pushed in `25afddb`. Both Marp steps are now `continue-on-error`, and a new `scripts/assemble-site.js` runs after them:

- system map and landing page are copied **unconditionally**;
- the deck card is kept **only if the deck was actually produced**, so the published site can never link to a page that was not built;
- a missing deck emits a workflow warning, not a failure.

The card is delimited in `docs/index.html` by `<!-- deck-card:start -->` / `<!-- deck-card:end -->`. The script exits non-zero if those markers go missing, rather than silently publishing a dangling link — `check-site.js` would fail on that anyway, which would put us right back where we started.

### Verified both branches

```
CASE A · deck built
  cards on landing: [The system map, Platform architecture deck]
  click "The system map"              -> /seans-mfe-tool/system-map.html
  click "Platform architecture deck"  -> /seans-mfe-tool/slides/platform-architecture.html
  map: 10 sections, 4 diagrams, 9 stages
  check-site: OK — 4 pages, 4 in-site links, 0 problems

CASE B · deck FAILED
  ::warning::Slide deck was not built — publishing the docs site without it.
  cards on landing: [The system map]
  click "The system map"              -> /seans-mfe-tool/system-map.html
  map: 10 sections, 4 diagrams, 9 stages   ← unaffected
  check-site: OK — 3 pages, 2 in-site links, 0 problems
```

A bad URL renders the landing page via `404.html` in both cases. Reproduce either locally:

```bash
# deck present
mkdir -p _site/slides && echo '<html><title>deck</title></html>' > _site/slides/platform-architecture.html
node scripts/assemble-site.js _site && node scripts/check-site.js _site

# deck build failed
rm -rf _site && node scripts/assemble-site.js _site && node scripts/check-site.js _site
```

---

## On the broken system-map link — I could not reproduce it, and here is why I think you saw it

**The live link works.** I mirrored the deployed bytes and clicked through in Chromium, starting both with and without the trailing slash:

| Start URL | Result |
|---|---|
| `…/seans-mfe-tool` | 301 → `…/seans-mfe-tool/`, card resolves to `…/seans-mfe-tool/system-map.html`, loads, 10 sections / 4 diagrams |
| `…/seans-mfe-tool/` | same |

No console errors, no failed requests. All five external links on the landing page return 200. The deployed `system-map.html` is structurally complete (one `<!doctype>`, one `</html>`, script block intact).

**What I think happened — timing.** From the workflow history:

| Time (UTC) | Event |
|---|---|
| ~04:00 | I told you the URL 404s. **It did.** |
| 04:03:50 | You merged #358 |
| **04:04:42** | First deploy completed — site goes live |

If you clicked the README link in that window, you got a genuine 404 — and both your browser and the Pages CDN cache those. A hard refresh (<kbd>Cmd/Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>R</kbd>) on <https://falese.github.io/seans-mfe-tool/system-map.html> should clear it.

**Worth knowing:** #358 merged only the first commit, so what is live right now is the **pre-terminology-pass** version — "factory", "control tower", "the brief". This PR is what replaces it.

If it is still broken for you after a hard refresh, tell me what you see — the URL in the bar and any console output — and I will keep digging. "Works in my browser" is not a fix, so I would rather chase a real reproduction than close this on my own testing.

---

